PR #2252, opened 2018-08-09
AMP-26020: Run unit tests in jenkins

developster:


jdeanquin-dg:
@developster please fix conflcts

PR #2305, opened 2018-09-03
 AMP-28075: Add resource type attribute for web link and file name validation

vchihai:


a-martirosyan:
related AMPOffline ticket is already merged
so there is no reason to keep this PR open

merging

PR #2333, opened 2018-09-06
Feature/amp 28175/new activity preview hf

dbianco:
Activity Preview feature: CR amp 28175


amp/TEMPLATE/reamp/modules/activity-preview/actions/CommonListsActions.jsx
```diff
@@ -0,0 +1,175 @@
+import commonListsApi from '../api/CommonListsApi';
+
+
+/**
+ * @author Daniel Oliva
```
jdeanquin-dg: Since the code is coming from amp, please remove the author and add the file and a link to the GitHub revision from there we took this code so it is traceable when we decide to merge both codes (preview from Offline and and online preview)

dbianco: The classnames have the same name as amponline (and most of the folder structure remains), but I couldn't use the code from it since it has lot of bugs and the endpoints that both apps consumes are different. Good luck if you try to merge the code, I will remove my name so nobody could find me at such moment.

PR #2338, opened 2018-09-07
Merge hotfix 3.2.1 into develop

jdeanquin-dg:


jdeanquin-dg:
we will use an intermediate branch since develop got updated and we cannot merge a hotfix branch with develop

PR #2349, opened 2018-09-14
Task/AMP-28307/preview settings endpoint

jdeanquin-dg:


jdeanquin-dg:
You are right, at least it should have gone to /amp/settings. I’ll move it there and I will create a ticket to move what needs to be moved to setting def and properly refactor /amp/settings.

jdeanquin-dg:
I will also fix add a ticket to move /public settings definition that shouldn’t have gone there too

PR #2352, opened 2018-09-14
Task/AMP-28307/add ws currency

jdeanquin-dg:


amp/WEB-INF/src/org/digijava/module/aim/util/CurrencyUtil.java
```diff
@@ -765,25 +770,40 @@ public static void checkDatabaseSanity(Session session) throws AimException {
...
+        if (tm != null && tm.getAppSettings() != null && tm.getAppSettings().getCurrencyId() != null) {
+            curr = CurrencyUtil.getAmpcurrency(tm.getAppSettings().getCurrencyId());
+        } else {
+            String currCode = FeaturesUtil.getGlobalSettingValue(GlobalSettingsConstants.BASE_CURRENCY);
+            if (currCode != null) {
+                curr = CurrencyUtil.getAmpcurrency(currCode);
+            }
+        }
+        return curr;
+    }
```
nmandrescu: This can be simplified to one line `CurrencyUtil.getWorkspaceCurrency(TeamUtil.getCurrentMember())`

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/settings/SettingsUtils.java
```diff
@@ -434,6 +436,15 @@ public static JsonBean getGeneralSettings() {
...
 
+    private static String  getDefaultCurrencyCode() {
+        //It should never be null since its coming from gs but just in case we default to USD
+        AmpCurrency curr = CurrencyUtil.getEffectiveCurrency();
+        if (curr != null) {
+            return curr.getCurrencyCode();
+        } else {
+            return CurrencyUtil.BASE_CODE;
+        }
+    }
```
nmandrescu: Should we name it `getEffectiveCurrencyCode` for clarity that it may be different from the default one from WS? Please move this to `CurrencyUtil` for reuse.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/settings/SettingsConstants.java
```diff
@@ -49,6 +49,8 @@
     public static final String CUSTOM = "custom";
 
     public static final String REORDER_FUNDING_ITEM_ID = "reorder-funding-item";
+    public static final String WORKSPACE_CURRENCY_CODE = "workspace-" + CURRENCY_ID;
```
nmandrescu: From what I see, this will be effective currency code, either for the one customized for the current ws or the global one including for public without ws. So may be to simply use `CURRENCY_ID` without this extra constant? Or at least a more general one like `effective-currency-code`

jdeanquin-dg:
check style will be solved in the main branch since the code reported has not been modified on scope of this ticket

PR #2353, opened 2018-09-16
Task/AMP-28318/activity info endpoint

jdeanquin-dg:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/InterchangeEndpoints.java
```diff
@@ -420,6 +420,18 @@ public JsonBean getProject(@PathParam("projectId") Long projectId, JsonBean filt
         return InterchangeUtils.getActivity(projectId, filter);
     }
 
+    @GET
+    @Path("/projects/info/{projectId}")
```
nmandrescu: If you don't mind, let's make the path a bit simpler & clearer `/info/{projectId}`

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityErrors.java
```diff
@@ -48,5 +48,8 @@
     public static final ApiErrorMessage FUNDING_PLEDGE_ORG_GROUP_MISMATCH = new ApiErrorMessage(30, 
             "The organization group of the pledge doesn't match with funding donor organization group");
     public static final ApiErrorMessage FIELD_TOO_MANY_VALUES_NOT_ALLOWED = new ApiErrorMessage(31, "Too many values");
+    public static final ApiErrorMessage ACTIVITY_NOT_FOUND = new ApiErrorMessage(32,
+            "Activity not found");
```
nmandrescu: Please rather use `AuthRule.VIEW_ACTIVITY` rule to report security or not found use case.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/InterchangeUtils.java
```diff
@@ -834,4 +838,28 @@ private static Field getField(Class<?> clazz, String fieldname) {
         
         return null;
     }
+
+    public static ActivityInformation getActivityInformation(Long projectId) {
+        AmpActivityVersion project = loadActivity(projectId);
+        if (project == null) {
+            //so far project will never be null since an exception will be thrown
+            //I leave the code prepared to throw the appropriate response code
+            ApiErrorResponse.reportResourceNotFound(ActivityErrors.ACTIVITY_NOT_FOUND);
```
nmandrescu: `AuthRule.VIEW_ACTIVITY` will validate that the activity is correct, so if it was validated, then here would be an abnormal case, i.e. `RuntimeException`

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/InterchangeEndpoints.java
jdeanquin-dg: done

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityErrors.java
jdeanquin-dg: the endpoint needs to be used for a non logged in user (coming from the portal, or directly from the link) so thats why I didn't use the VIEW_ACTIVITY AuthRule

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/InterchangeUtils.java
jdeanquin-dg: Please see comment above

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityErrors.java
nmandrescu: Sounds good then, 404 would clarify. Please note however that VIEW_ACTIVITY does multiple things that now you need to do else where:
1. Activity exists (that you did)
2. Activity is the latest version. Where will the redirect to the latest be handled? E.g. an AMP filter can do that or in info EP response we can provide the latest and UI will redirect.
3. I can see currently private activities can be seen by public users (e.g. shared linked from someone else or logged out). If it is a bug, then we will need a rule to check for private activities as well.

jdeanquin-dg: for 2 I will add extra info saying its not the last version since currently the preview(old) if you see a version that is not the last, it says *Requested activity version was not the latest version. Preview switched to showing the last version!* you refresh and it never shows the last version. I will ask dani slightly modify the text and add link so a the new version is called.

Regarding private ws you are right, it does not do  the check, but in the current preview you are able to see the private activity if you are not logged in. You get the message but you see it anyways. Ill check with Vanessa if we should fix now , or as an improvement.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityEPConstants.java
```diff
@@ -57,6 +57,8 @@
             InterchangeUtils.underscorify(ActivityFieldsConstants.MODIFIED_BY);
 
     public static final String AMP_ACTIVITY_ID = "amp-activity-id";
+    public static final String AMP_ACTIVITY_LAAST_VERSION_ID = "amp-activity-last-version-id";
```
nmandrescu: pls fix small typo for LAAST :)

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityInformation.java
```diff
@@ -10,6 +10,9 @@
     @JsonProperty(ActivityEPConstants.AMP_ACTIVITY_ID)
     private Long ampActivityId;
 
+    @JsonProperty(ActivityEPConstants.AMP_ACTIVITY_LAAST_VERSION_ID)
+    private Long ampASctiviylastVersionId;
```
nmandrescu: pls fix type in `ampASctiviylastVersionId` to `ampActiviyLastVersionId`

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityEPConstants.java
jdeanquin-dg: lol done

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityInformation.java
jdeanquin-dg: done

PR #2354, opened 2018-09-17
AMP-28311: Add calendar EP and sync up support

vchihai:


amp/WEB-INF/src/org/digijava/module/aim/dbentity/AmpFiscalCalendar.java
```diff
@@ -23,25 +24,27 @@
 
 public class AmpFiscalCalendar implements Serializable, Identifiable, OrgProfileValue, CalendarConverter {
     
-    //IATI-check: to be ignored
-//  @Interchangeable(fieldTitle="Fiscal Calendar ID")
+    @JsonProperty("id")
     private Long ampFiscalCalId;
-//  @Interchangeable(fieldTitle="Start Month Number")
+    
+    @JsonProperty("start_month_num")
```
nmandrescu: Please use kebab-case for json properties in API for new fields (and in general when adding new fields)

PR #2356, opened 2018-09-18
AMP-28314: Create endpoint for fetching funding information for activity

vchihai:
preview

### DIFF
--- a/amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/preview/AmountSerializer.java
+++ b/amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/preview/AmountSerializer.java
@@ -1,0 +1,25 @@
+package org.digijava.kernel.ampapi.endpoints.activity.preview;
+
+import java.io.IOException;
+import java.math.BigDecimal;
+
+import org.codehaus.jackson.JsonGenerator;
+import org.codehaus.jackson.JsonProcessingException;
+import org.codehaus.jackson.map.JsonSerializer;
+import org.codehaus.jackson.map.SerializerProvider;
+
+/**
+ * Amount json serializer used for activity preview endpoint
+ * 
+ * @author Viorel Chihai
+ *
+ */
+public class AmountSerializer extends JsonSerializer<Double>  {
+
+    @Override
+    public void serialize(Double value, JsonGenerator jgen, SerializerProvider provider)
+            throws IOException, JsonProcessingException {
+        jgen.writeString(BigDecimal.valueOf(value).setScale(0, BigDecimal.ROUND_HALF_UP).toString());        
+    }
+
+}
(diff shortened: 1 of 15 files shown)


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/preview/AmountSerializer.java
```diff
@@ -0,0 +1,25 @@
...
+ * 
+ * @author Viorel Chihai
+ *
+ */
+public class AmountSerializer extends JsonSerializer<Double>  {
+
+    @Override
+    public void serialize(Double value, JsonGenerator jgen, SerializerProvider provider)
+            throws IOException, JsonProcessingException {
+        jgen.writeString(BigDecimal.valueOf(value).setScale(0, BigDecimal.ROUND_HALF_UP).toString());
```
nmandrescu: I checked for usages and I'd like to confirm: should the amounts like PPC/RPC, subtotals be rounded up to the decimal point in preview? I understand normally we don't provide decimal point entries in the orig. currency and it may be reasonable to do the same for converted amounts, however AMP uses amount formats that allows decimal points and for values like `undisbursed_balance` it may make sense, so just want to be sure this rounding is used correctly.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/preview/PreviewActivityService.java
```diff
@@ -0,0 +1,290 @@
...
+            AmpActivityVersion activity = ActivityUtil.loadActivity(activityId);
+            
+            AmpCurrency currency = null;
+            if (currencyId != null) {
+                currency = CurrencyUtil.getAmpcurrency(currencyId);
+            }
+            
+            if (currencyId == null || currency == null) {
+                throw new ApiRuntimeException(Response.Status.BAD_REQUEST, ApiError.toError(
+                        "Parameter 'currency-id' is not specified or it is not a valid value"));
```
nmandrescu: Please also check input for right `activityId` and define constants to use as a parameter name in EP and here in error details to ensure consistency.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/preview/PreviewActivityService.java
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    private Double convertAmount(Double originalAmount, String fromCurrencyCode, String toCurrencyCode,
+            LocalDate transactionDate) {
+
+        return convertAmount(originalAmount, fromCurrencyCode, toCurrencyCode, null, transactionDate);
+    }
+
+    private Double convertAmount(Double originalAmount, String fromCurrencyCode, String toCurrencyCode,
+            Double fixedExchangeRate, LocalDate transactionDate) {
```
nmandrescu: It could be helpful to move `convertAmount` methods for reuse at AMP level

PR #2381, opened 2018-09-27
AMP-28351: Fix visibility of source role in activity api

vchihai:


jdeanquin-dg:
Pr tested with @mbaralo

PR #2387, opened 2018-09-28
AMP-28264: Update name for 'show empty row in reports' gs

vchihai:


jdeanquin-dg:
its not needed in 2.13

PR #2399, opened 2018-10-02
Task/amp 28399/js vulnerabilities filters

vchihai:


jdeanquin-dg (approved):

jdeanquin-dg:
please test that filters and GIS loads correctly, just q quick round

PR #2414, opened 2018-10-08
AMP-28417: Fix saving new fundings with mtef transactions

vchihai:


amp/WEB-INF/src/org/dgfoundation/amp/onepager/util/ActivityUtil.java
```diff
@@ -230,6 +231,15 @@ public static AmpActivityVersion saveActivityNewVersion(AmpActivityVersion a,
                 }
             }
 
+            for (AmpFunding f : af) {
```
jdeanquin-dg: please check that we are already iterating amp fundings in line 220. lets try to make updateFundingDetails generic so we can update the ids there and we avoid iterating a second time

vchihai: There are 2 iterations - one is used to filter transactions without values (like 0) and the second one used for setting the funding detail. I can refactor more code to avoid double iteration.

jdeanquin-dg: yes please, lets try to refactor so we iterate only once

PR #2420, opened 2018-10-09
AMP-28275: Fix show funding for non-logged in users

vchihai:


vchihai:
@jdeanquin-dg , you are right. Closing and creating a new one against hotfix/v3.2.2

PR #2421, opened 2018-10-09
AMP-28275: Fix show funding for non-logged in users (dev)

vchihai:


jdeanquin-dg:
won't merge so far, keeping the branch just in case

PR #2428, opened 2018-10-11
AMP-28342: Fix team filter for sync endpoint

vchihai:


amp/WEB-INF/src/org/dgfoundation/amp/ar/AmpARFilter.java
```diff
@@ -878,19 +880,19 @@ public void fillWithDefaultsFilter(Long ampReportId)
...
-        if (request != null)
-            tm = (TeamMember) request.getSession().getAttribute(Constants.CURRENT_MEMBER);
-        
-        if (tm == null || tm.getTeamId() == null )
-            tm  = null;
 
-        if (tm != null) {
+        if (teamMember == null && request != null) {
+            teamMember = (TeamMember) request.getSession().getAttribute(Constants.CURRENT_MEMBER);
+        }
```
nmandrescu: I am wondering if `teamMember` wasn't stored within `AmpARFilter` on purpose. E.g. if user logs out and refreshes a report, will it work ok? Or could there be any other "cached" filter issue?

vchihai: AmpARFilter object is stored in the user session. If the user logs out, the object will not be available anymore.

nmandrescu: will it update correctly when switching workspace?

vchihai: Yes, you are right. It is recreated every time when the user switches to a  new workspace

PR #2444, opened 2018-10-22
AMP-25480: Update activit modified date when related entity are changed

vchihai:


amp/WEB-INF/src/org/digijava/module/aim/dbentity/AmpActivityFields.java
```diff
@@ -2172,5 +2175,17 @@ public void addCostAmount(AmpFundingAmount costAmount) {
             this.costAmounts.add(costAmount);
         }
 
+    @Override
+    public void touch() {
+        Date updateDate = Calendar.getInstance().getTime();
+        setUpdatedDate(updateDate);
+        setModifiedDate(updateDate);
+
+        AmpTeamMember currentAmpTeamMember = TeamMemberUtil.getCurrentAmpTeamMember(TLSUtils.getRequest());
```
developster: There is an assumption that teamMember passed to org.dgfoundation.amp.onepager.util.ActivityUtil#saveActivityNewVersion coincides with the one from current http request. Also activities could be saved outside of http request scope.

developster: For example in AMP Offline one user can save an activity on behalf of other user.
Also, we might need the possibility to save many activities on behalf on many different users inside same http request and/or tranzaction.

amp/WEB-INF/src/org/digijava/module/message/jobs/ActivityAutomaticValidationJob.java
```diff
@@ -40,7 +40,6 @@
      */
     protected AmpActivityVersion validateActivity(Session session, AmpTeamMember member, AmpActivityVersion oldActivity) throws CloneNotSupportedException {
         AmpActivityVersion prevVersion = oldActivity.getAmpActivityGroup().getAmpActivityLastVersion();
-        oldActivity.setModifiedDate(Calendar.getInstance().getTime());
         oldActivity.setModifiedBy(member);
```
developster: If interceptor is working correctly then we don't have to call setModifiedBy() anymore. Right? We have to remove all calls to setModifedBy and setModifiedDate.

amp/WEB-INF/src/org/digijava/module/aim/audit/AuditActivityInfo.java
```diff
@@ -0,0 +1,34 @@
+package org.digijava.module.aim.audit;
+
+import org.digijava.module.aim.dbentity.AmpTeamMember;
+
+public class AuditActivityInfo {
```
developster: Please add here method that encapsulates try/finally logic to avoid code repetition and misuse of this class.
Also remove getters/setter to field & static variable and expose only `runInContext()` and `getTeamMember()` methods.

PR #2445, opened 2018-10-23
AMP-28370: Update uglify-js version for backbone localStorage module in tabs

vchihai:


jdeanquin-dg:
Please Test tabs separate before merging

PR #2458, opened 2018-10-29
AMP-28105: Change relationship type between activity and structures

vchihai:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/gis/services/LocationService.java
```diff
@@ -409,14 +409,7 @@ public static FeatureGeoJSON buildFeatureGeoJSON(AmpStructure structure) {
...
             
-            Set<AmpActivityVersion> av = structure.getActivities();
-            List<Long> actIds = new ArrayList<Long>();
-
-            for (AmpActivityVersion ampActivity : av) {
-                actIds.add(ampActivity.getAmpActivityId());
-            }
-
-            fgj.properties.put("activity", new POJONode(actIds));
+            fgj.properties.put("activity", new POJONode(structure.getActivity()));
```
developster: Will GIS work after list of ids was changed to just one id?

PR #2459, opened 2018-10-29
AMP-26948: Refresh validation for location when the country is added automatically

vchihai:


amp/WEB-INF/src/org/dgfoundation/amp/onepager/components/features/tables/AmpLocationFormTableFeature.java
```diff
@@ -353,4 +354,19 @@ public void locationSelected(AmpCategoryValueLocations choice, IModel<AmpActivit
         }
         set.add(activityLocation);
     }
+
+    public void reloadValidationFields(AjaxRequestTarget target) {
+        ComponentHierarchyIterator componentIterator = this.visitChildren();
```
jdeanquin-dg: this.visitChildren() is deprecated (dont know if in 6.18.0 version but in 6.30.0 for sure it will be
you can implement it with a lamba function using an IVisitor and you can avoid using instance of
`
this.visitChildren(AmpCollectionValidatorField.class, (IVisitor<AmpCollectionValidatorField, Void>) (component, visit) -> component.reloadValidationField(target));`

PR #2462, opened 2018-10-30
AMP-28487: Fix js issue when adding a transaction

vchihai:


amp/WEB-INF/src/org/dgfoundation/amp/onepager/components/QuarterInformationPanel.java
```diff
@@ -33,7 +33,7 @@
  * @since 04 Feb 2013
  */
 public class QuarterInformationPanel extends AmpFieldPanel {
-    private static final JavaScriptResourceReference JS_FILE = new JavaScriptResourceReference(
+    public static final JavaScriptResourceReference JS_FILE = new JavaScriptResourceReference(
             QuarterInformationPanel.class, "QuarterInformationPanel.js");
```
jdeanquin-dg: since the file has already been added on AmpSubsectionFeatureFundingPanel#renderHead I think we can remove that line from here

PR #2464, opened 2018-10-30
AMP-27457: Update wicket version to 6.29.0

vchihai:


jdeanquin-dg:
merge only when all substasks were tested and merged separately

PR #2476, opened 2018-11-01
AMP-28507: Fix checking if disaster response is required or not

vchihai:


amp/WEB-INF/src/org/digijava/module/aim/dbentity/AmpFundingDetail.java
```diff
@@ -218,7 +218,6 @@ public void setExpenditureClass(final AmpCategoryValue expenditureClass) {
     private Long checkSum;
     
     @Interchangeable(fieldTitle = "Disaster Response", importable = true, required = REQUIRED_ALWAYS,
-            fmPath = FMVisibility.ANY_FM + ActivityEPConstants.COMMITMENTS_DISASTER_RESPONSE_FM_PATH + "|" + ActivityEPConstants.DISBURSEMENTS_DISASTER_RESPONSE_FM_PATH,
```
nmandrescu: We still need to check FM paths to decide if to provide in fields def EP or not.

PR #2507, opened 2018-11-15
AMP-28569

developster:
- Added a method to delete all unused visibility rules (these rules prevented FM entry deletion)
- Enabled FK on amp_menu_entry_view.rule_id to prevent accidental deletes


amp/xmlpatches/3.2.00/AMP-25234-remove-saiku.xml
```diff
@@ -7,8 +7,16 @@
     <author>Octavian Ciubotaru</author>
     <description>Remove saiku from menus and FM.</description>
     <apply>
+        <script>
+            <lang delimiter=";" type="postgres"><![CDATA[
+                ALTER TABLE amp_menu_entry_view
+                    ADD CONSTRAINT amp_menu_entry_view_amp_visibility_rule_id_fk
+                    FOREIGN KEY (rule_id) REFERENCES amp_visibility_rule (id);
+            ]]></lang>
+        </script>
```
nmandrescu: Please define a separate patch, since it won't rerun on DBs that already executed the current patch successfully.

amp/xmlpatches/3.2.00/AMP-25234-remove-saiku.xml
```diff
@@ -7,8 +7,16 @@
...
         <script>
             <lang delimiter=";" type="bsh"><![CDATA[
+                org.digijava.kernel.xmlpatches.MenuRemover.removeUnusedVisibilityRules();
```
nmandrescu: Probably we can also duplicate this call into the separate patch.

PR #2515, opened 2018-11-16
AMP-28572: Sync possible values fields

vchihai:


amp/WEB-INF/src/org/digijava/kernel/services/sync/SyncService.java
```diff
@@ -245,6 +263,10 @@ private boolean ratesChanged(Date lastSyncTime) {
         }
         return fieldFilter;
     }
+    
+    private List<String> findAllActivityPossibleValues() {
```
developster: Please rename to `findAllActivityFieldsWithPossibleValues`.

PR #2531, opened 2018-11-21
AMP-28590: Fix showing sectors in activity preview

vchihai:


amp/WEB-INF/src/org/digijava/module/aim/action/EditActivity.java
```diff
@@ -1178,12 +1178,9 @@ public ActionForward execute(ActionMapping mapping, ActionForm form,
           eaForm.getDocuments().setReferenceDocs(null);
 
           eaForm=setSectorsToForm(eaForm, activity);
-            if (isPreview) {
```
jdeanquin-dg: Actually I think the proper fix should be to change how isPreview is calculated (Revert to previous code). The hack we are using to call the new preview is not using isPreview

PR #2543, opened 2018-11-22
AMP-28275: Fix show funding for non-logged in users

jdeanquin-dg:


jdeanquin-dg:
Wont merge

jdeanquin-dg:
keeping the branch just in case

PR #2549, opened 2018-11-23
Feature/amp 27526/pledget filters

ginchauspe:


mbaralo:
Please @jdeanquin-dg when validating FE please also validate the related BE and modify the tickets status under https://jira.developmentgateway.org/browse/AMP-24754.Thanks!

PR #2552, opened 2018-11-26
AMP-26025: Approval status as enum

developster:


amp/WEB-INF/src/org/digijava/module/aim/helper/Constants.java
```diff
@@ -197,6 +194,10 @@ private Constants() {
     public static final String ACTIVITY_STATUS_PROPOSED         = "Proposed";
     public static final String ACTIVITY_STATUS_CONSIDERED       = "Considered";
     
+    public static final Set<ApprovalStatus> ACTIVITY_NEEDS_APPROVAL_STATUS_SET = ImmutableSet.of(
+            ApprovalStatus.STARTED_STATUS,
+            ApprovalStatus.EDITED_STATUS,
+            ApprovalStatus.REJECTED_STATUS);
     public static final String ACTIVITY_NEEDS_APPROVAL_STATUS   = "'created', 'started', 'edited', 'rejected'";
```
nmandrescu: Can we remove this? I saw you replaced it in some places with `ACTIVITY_NEEDS_APPROVAL_STATUS_SET`. On the other hand I see we no longer have 'created' which is present here - could be a leftover.
cc @jdeanquin-dg

amp/WEB-INF/src/org/digijava/module/aim/dbentity/ApprovalStatusType.java
```diff
@@ -0,0 +1,77 @@
...
+
+import org.hibernate.HibernateException;
+import org.hibernate.engine.spi.SessionImplementor;
+import org.hibernate.type.StringType;
+import org.hibernate.usertype.UserType;
+
+/**
+ * @author Octavian Ciubotaru
+ */
+public class ApprovalStatusType implements UserType {
```
nmandrescu: I think you can do the same by using 'org.hibernate.type.EnumType' in .hbm.xml file (e.g. https://github.com/devgateway/amp/blob/a93e1fe146b66eb2c2933968abd9d694759b626c/amp/WEB-INF/src/org/digijava/module/aim/dbentity/AmpMenuEntry.hbm.xml#L63) and configure "useNamed" to true to keep strings.

nmandrescu:
Please also fix checkstyle

nmandrescu:
@jdeanquin-dg please also review, since touches form/preview code

amp/WEB-INF/src/org/digijava/module/aim/dbentity/ApprovalStatusType.java
developster: Using EnumType will require changing `startedapproved` to `STARTED_APPROVED_STATUS` in database. I wanted to avoid renaming to prevent possible bugs coming out of this.

nmandrescu: doesn't it get the .name that you can configure to return the right value?

amp/WEB-INF/src/org/digijava/module/aim/helper/Constants.java
developster: I forgot to flag this in daily. Actually there are 4 countries that do use 'created' value. If I remove created value from enum those activities will be impossible to read through hibernate.
And actually it is an error that I excluded CREATED from the _SET variable. Adding it back.
From the same category is 'rejected' status which was never used. None of the countries use it. Will report a ticket to tackle both statuses which will probably have to be removed.

developster: I made an error and meant 'not_approved' instead of 'rejected'. 'not_approved' is somewhat misleading as it could be represented by 'rejected'/'started'/'edited', basically any status except 'approved' and 'startedapproved'. So we should definitely remove it. On the other hand, 'rejected' is in use by many countries and we'll have to keep it.

@jdeanquin-dg , I am not sure what you mean by approval_status.

The original `ACTIVITY_NEEDS_APPROVAL_STATUS` constant is used for building several queries and it makes sense to keep it as is for now. The set was used for the rest of the cases where we had code like this `if (ACTIVITY_NEEDS_APPROVAL_STATUS.contains("edited")) {...}`.

amp/WEB-INF/src/org/digijava/module/aim/dbentity/ApprovalStatusType.java
developster: Could you please elaborate more on that? From what I see there is no way to do this.

developster:
This PR has checkstyle violations in `WorkspaceFilter`. There is no real need to modify this class because we have another PR where we modified this class and will have merge conflicts.

amp/WEB-INF/src/org/digijava/module/aim/helper/Constants.java
developster: Updated an existing ticket to cleanup approval status. See AMP-26951.

amp/WEB-INF/src/org/digijava/module/aim/dbentity/ApprovalStatusType.java
nmandrescu: My bad memories about code level only hacks - not applicable in this case. Though it would be nice to avoid this class and may be indeed use a patch to convert DB entries?

Btw I saw org.hibernate.type.EnumType already implements UserType so could be simpler to extend it (with `nullSafeGet` and `nullSafeSet` only)?

nmandrescu:
Thanks for fixing checkstyle violations. Sure, the remaining ones from `WorkspaceFilter` can be deferred to the other PR.

amp/WEB-INF/src/org/digijava/module/aim/helper/Constants.java
jdeanquin-dg: sorry for. approval_status I meant created

amp/WEB-INF/src/org/digijava/module/aim/dbentity/ApprovalStatusType.java
jdeanquin-dg: I was going to suggest the same, to override EnumType .that we can also do it with AmpOfflineVersionType that follows the same logic, and we could have smaller classes

developster: Re-implemented enum as a subclass of EnumType. AmpOfflineVersionType was left untouched because AmpOfflineVersion is not an enum.

jdeanquin-dg: right my bad, I overlooked it

amp/WEB-INF/src/org/digijava/module/aim/dbentity/ApprovalStatus.java
```diff
@@ -0,0 +1,56 @@
...
+    CREATED("created", 0),
+    APPROVED_STATUS("approved", 1),
+    EDITED_STATUS("edited", 2),
+    STARTED_APPROVED_STATUS("startedapproved", 3),
+    STARTED_STATUS("started", 4),
+    /**
+     * Not in use. Synonym to rejected?
+     */
+    NOT_APPROVED("not_approved", 5),
+    REJECTED_STATUS("rejected", 6);
```
nmandrescu: I was looking at `ApprovalStatus` to replicate in AMP Offline and decided to skip adding `_STATUS` suffix there. What do you think - should we catch the moment to simplify the name in AMP as well since the enum itself already explains that this about status?

PR #2555, opened 2018-11-27
AMP-28529: Remove JsonBean in filters and scorecards EP

developster:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/FiltersEndpoint.java
```diff
@@ -74,14 +78,13 @@
     private static final String SECTORS_SUFFIX = " Sectors";
     private static final Logger logger = Logger.getLogger(FiltersEndpoint.class);
     
-    /** the value to use as a filter value when filtering booleans for ANY DEFINED */
-    public static final String ANY_BOOLEAN = "999888777";
```
vchihai: Any defined means TRUE or FALSE (not null values)
Did you check this scenario while removing this?

PR #2563, opened 2018-11-30
AMP-25570: Sync field changes

developster:


amp/WEB-INF/src/org/digijava/kernel/services/sync/model/SystemDiff.java
```diff
@@ -69,6 +69,11 @@
     @JsonProperty("feature-manager")
     private boolean featureManager;
 
+    /**
+     * Set when field definitions any entity changed.
+     */
+    private boolean fields;
```
nmandrescu: I think it's better to name 'activity-fields' to be clear and also later use another ones for contacts and resources.

PR #2567, opened 2018-12-05
AMP-28536: Remove JsonBean from data and gis EP

developster:
Also:
- Fixed Export Project Sites and Export Administrative Locations from GIS
- Fixed error reporting in API when Accept header is a list of mime types


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/reports/ReportConfig.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModelProperty;
+
+/**
+ * @author Octavian Ciubotaru
+ */
+public class ReportConfig {
+
+    @JsonProperty("layers-view")
```
developster: @nmandrescu I couldn't find anywhere this in use. Maybe we should drop it?

PR #2584, opened 2018-12-11
Task/ampet 140/comparision service

MistreW:


amp/TEMPLATE/ampTemplate/site-config.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <!DOCTYPE page [
-<!ENTITY ampVersion "${project.version}">
+<!ENTITY ampVersion "2.13.5-SNAPSHOT">
```
jdeanquin-dg: this doesn't need to be changed. in 2.13.5 this is replaced by the build process

amp/WEB-INF/src/org/digijava/module/aim/util/ActivityVersionUtil.java
```diff
@@ -70,13 +71,15 @@ public static boolean implementsVersionable(Class[] array) {
...
-        return generateFormattedOutput(request, out, null);
+    public static String generateFormattedOutput(Output out) throws WorkerException {
+        return generateFormattedOutput(out, null);
     }
 
-    public static String generateFormattedOutput(HttpServletRequest request, Output out, Output out1) throws WorkerException {
-        Site site = RequestUtils.getSite(request);
-        String langCode = RequestUtils.getNavigationLanguage(request).getCode();
+    public static String generateFormattedOutput(Output out, Output out1) throws WorkerException {
+        // Site site = RequestUtils.getSite(request);
```
jdeanquin-dg: we can avoid commenting code so its easier to spot changes. Also if we need to reference the changed code we can use git history for this

amp/WEB-INF/src/org/digijava/module/aim/annotations/activityversioning/ActivityVersioningService.java
```diff
@@ -0,0 +1,411 @@
+package org.digijava.module.aim.annotations.activityversioning;
```
jdeanquin-dg: we can add the code to ActivityVersioningUtil since we already have methods there that does related functionality . We can create a ticket to split that class in order to make it smaller

amp/TEMPLATE/ampTemplate/site-config.xml
jdeanquin-dg: This needs to be resolved before merging or approving the ticket

jdeanquin-dg:
code already merged

PR #2588, opened 2018-12-12
extract comparsion code

Rahelalemayehu:


jdeanquin-dg:
The pr should have at list the name of the ticket that is implementing this change so then Jira can match and you can see it from the ticket

amp/WEB-INF/src/org/digijava/module/aim/util/ActivityVersionUtil.java
```diff
@@ -11,7 +12,7 @@
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
-
+//import org.digijava.kernel.request.TLSUtils;
```
jdeanquin-dg: if an import is not in used or duplicate in order to keep code clean we should remove it instead of commenting. we can trace changes using GitHub history

amp/WEB-INF/src/org/digijava/module/aim/util/ActivityVersionUtil.java
```diff
@@ -70,14 +71,20 @@ public static boolean implementsVersionable(Class[] array) {
         return ret;
     }
 
-    public static String generateFormattedOutput(HttpServletRequest request, Output out) throws WorkerException {
-        return generateFormattedOutput(request, out, null);
+    //public static String generateFormattedOutput(HttpServletRequest request, Output out) throws WorkerException {
```
jdeanquin-dg: same as before, when we are changing a method is difficult to trace down what has changed with commented code, if you see what has changed in this pr you can notice. So the best practice is to just change the code that needs to be changed, and if you need to

amp/WEB-INF/src/org/digijava/module/aim/annotations/activityversioning/ActivityVersioningService.java
```diff
@@ -0,0 +1,366 @@
+package org.digijava.module.aim.annotations.activityversioning;
```
jdeanquin-dg: since we already have a util class for activity versioning we could have used that class, as I did. we also have methods related to comparison in that class. then we would have to create a ticket as followup to provide a proper separation of classes, based on functionality

Rahelalemayehu:
AMPET-140 updated

Rahelalemayehu:
AMPET-140 updated

jdeanquin-dg:
Closing since #2661 has already been merged

PR #2601, opened 2018-12-14
Feature/amp 28680/call ep

ginchauspe:


jdeanquin-dg:
this has been replaced by #2632

PR #2604, opened 2018-12-17
AMPET-140: Refactor/extract activity comparison code from struts action

Wessi:
The new branch contains a service method that is extracted from existing one to compare activity versions.

### DIFF
--- a/amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
+++ b/amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
@@ -65,7 +65,7 @@ public class CompareActivityVersions extends DispatchAction {
         }
 
         vForm.setOutputCollection(new ArrayList<CompareOutput>());
-        vForm.setOutputCollectionGrouped(ActivityVersionUtil.compareActivities(vForm.getActivityOneId(), vForm.getActivityTwoId()));
+        vForm.setOutputCollectionGrouped(ActivityVersionUtil.compareWithPrevious(vForm.getActivityOneId()));
 
         return mapping.findForward("forward");
     }
@@ -74,7 +74,6 @@ public class CompareActivityVersions extends DispatchAction {
                                 HttpServletResponse response) throws Exception {
         return mapping.findForward("reload");
     }
-
 
     private void setAdvancemode(CompareActivityVersionsForm vForm, HttpServletRequest request) {
         boolean ispartofamanagetmentworkspace = false;
--- a/amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
+++ b/amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
@@ -1,87 +1,52 @@
 package org.digijava.module.aim.action;
 
-import java.lang.reflect.Field;
-import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Calendar;
-import java.util.Collection;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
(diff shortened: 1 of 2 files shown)


Wessi:
I accept "you shouldn't change or format code you are not touching. We used GitHub in order to track changes, and to see why a line of code has changed". If I'm going to remove unused line of commented codes and imports line by line, that would be time consuming and tedious. And that's why most editors incorporated auto code indenting and formatting options. If we are refactoring an existing code we should also be cleaning and optimizing the readability of the code which we are touching.

amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
```diff
@@ -95,643 +60,35 @@ public ActionForward compare(ActionMapping mapping, ActionForm form, HttpServlet
...
-
-        }
-
-        return retVal;
-    }
-
-    public ActionForward enableMerge(ActionMapping mapping, ActionForm form, HttpServletRequest request,
-            HttpServletResponse response) throws Exception {
+        vForm.setOutputCollection(new ArrayList<CompareOutput>());
+        vForm.setOutputCollectionGrouped(ActivityVersionUtil.compareWithPrevious(vForm.getActivityOneId()));
```
jdeanquin-dg: this shouldn't have been put here since we will use this method from the audit logger, and activity comparison from activity preview should remain the same, comparing what we have as parameter. as a test is fine (if you want to test the code) but we should not commit this since its changing the functionality that its used from preview

Wessi: That's to test whether the code works or not... we can also remove it when we migrate the code to the audit logger... I want to know your feedback on the code that compares with its previous version.

Wessi: I don't still get any significant or insignificant wrong snippet on my code. It just worked dramatically without any further step to dig... hope you can debug and check or I'm not understanding you...

jdeanquin-dg: Sorry I didn't see the change, somehow GitHub didn't show. Ill provide my feedback now. However @yosephtesfaye code has already been merged so you can continue. But I think its valuable for everybody to provide feedback on every task.

As a general rule we don't commit code that then will be removed and we will reject commits with such code. If we need to add it to test, we can do it locally and then commit the code without test code. Its very difficult to track down code that will be removed or changed afterwards and with this you are changing the code that is working now.

amp/WEB-INF/src/org/digijava/module/aim/util/ActivityVersionUtil.java
```diff
@@ -1,42 +1,406 @@
...
+        return ActivityVersionUtil.groupOutputCollection(outputCollection);
+    }
+
+    public static Map<String, List<CompareOutput>> compareWithPrevious(Long activityOneId) throws Exception {
+        // Obtain previous activity id if(index<0) { //there is no previous version // }
+        AmpActivityVersion ampActivityOne = (AmpActivityVersion) PersistenceManager.getCurrentSession()
+                .load(AmpActivityVersion.class, activityOneId);
+
+        AmpActivityGroup group = ampActivityOne.getAmpActivityGroup();
+        Set<AmpActivityVersion> activities = group.getActivities();
```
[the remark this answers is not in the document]
Wessi: I have already done that... please see the below section of the code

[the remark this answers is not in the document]
jdeanquin-dg: yes I mean that you could have used a custom comparator to sort directly the set instead of adding the ids to the array list (You should have created an ArrayList
 of of the set to do so) , and sort the array list. Also if you sort it in reverse order you just look for the second element, if exist.

jdeanquin-dg:
Closing PR since #2681 has been merged

PR #2605, opened 2018-12-18
Task/ampet 141/compare to previous

MistreW:


amp/WEB-INF/src/org/digijava/module/aim/util/ActivityVersionUtil.java
```diff
@@ -327,4 +337,396 @@ public static AmpActivityVersion initializeActivity(AmpActivityVersion act) thro
...
+ 	                    output.setDescriptionOutput(auxAnnotation.fieldTitle());
+ 	                    outputCollection.add(output);
+ 	                }
+ 	            }
+ 	        }
+
+ 	        return groupOutputCollection(outputCollection);
+ 	    }
+ 	    
+ 	   public static void compareToPrevious(Long activityOneId) throws Exception {
```
jdeanquin-dg: This method will need to be called from a struts action. so it needs to return a value that will further be used by this struts action to render the comparison

PR #2615, opened 2018-12-19
AMP-28693: Added support for 5th admin level in filters and reports

developster:


amp/WEB-INF/src/org/dgfoundation/amp/ar/ColumnConstants.java
```diff
@@ -74,6 +74,7 @@
     public static final String CURRENT_COMPLETION_DATE_COMMENTS = "Current Completion Date Comments";
     public static final String DESCRIPTION_OF_COMPONENT_FUNDING = "Description of Component Funding";
     public static final String DISTRICT = "District";
+    public static final String COMMUNAL_SECTION = "Communal Section";
```
vchihai: In other DB this level is named as 'Commune'. Shouldn't we have a generic option for this?

PR #2650, opened 2019-01-02
AMP-26970: Removed conversion to pledge columns from AmpARFilterConverter

developster:


amp/WEB-INF/src/org/dgfoundation/amp/reports/converters/AmpARFilterConverter.java
```diff
@@ -280,16 +275,16 @@ private String findSubSectorColumnName(String columnName, AmpSector sector) {
         }           
         return res;
     }
-    
+
+    // TODO, talk with Viorel on this issue?
```
vchihai: shouldn't this comment be removed?

developster: :)

PR #2652, opened 2019-01-02
Task/ampet 142/changed-activities-list

MistreW:


amp/WEB-INF/src/org/digijava/module/aim/util/ActivityVersionUtil.java
```diff
@@ -346,6 +346,16 @@ public static AmpActivityVersion initializeActivity(AmpActivityVersion act) thro
         return act;
     }
     
+    public static Map<Long, Map<String, List<CompareOutput>>> compareActivities(Long[] ActivitiesId) throws Exception {
+        
+        Session session = PersistenceManager.getCurrentSession();
```
jdeanquin-dg: you don't need this session here

amp/WEB-INF/src/org/digijava/module/aim/util/ActivityVersionUtil.java
```diff
@@ -346,6 +346,16 @@ public static AmpActivityVersion initializeActivity(AmpActivityVersion act) thro
...
+        Map<Long, Map<String, List<CompareOutput>>> activitiesCollection = new HashMap<Long, Map<String, List<CompareOutput>>>();
```
jdeanquin-dg: For further development its good you used a map here

MistreW: what do you mean?

jdeanquin-dg: That using a map of longs instead of a list is good choice

amp/WEB-INF/src/org/digijava/module/aim/util/ActivityVersionUtil.java
```diff
@@ -346,13 +346,19 @@ public static AmpActivityVersion initializeActivity(AmpActivityVersion act) thro
...
-    public static Map<Long, Map<String, List<CompareOutput>>> compareActivities(Long[] ActivitiesId) throws Exception {
+    public static Map<Long, Map<String, List<CompareOutput>>> compareActivities(ArrayList<Long> ActivitiesId) throws Exception {
         
-        Session session = PersistenceManager.getCurrentSession();
         Map<Long, Map<String, List<CompareOutput>>> activitiesCollection = new HashMap<Long, Map<String, List<CompareOutput>>>();
-        for (int i=0; i < ActivitiesId.length; i++) {
-            Map<String, List<CompareOutput>> ActivityList = compareActivities(ActivitiesId[i]);
-            activitiesCollection.put(ActivitiesId[i], ActivityList);
+        for (int i=0; i < ActivitiesId.size(); i++) {
+            Session session = PersistenceManager.getCurrentSession();
```
jdeanquin-dg: I missed this in my previous review you already have the method that compares one activity with its previous version, we shall reuse that so we dont have duplicate code

jdeanquin-dg:
also @MistreW please fix checkstly violations

amp/WEB-INF/src/org/digijava/module/aim/util/ActivityVersionUtil.java
```diff
@@ -346,6 +346,16 @@ public static AmpActivityVersion initializeActivity(AmpActivityVersion act) thro
         return act;
     }
     
+    public static Map<Long, Map<String, List<CompareOutput>>> compareActivities(Long[] ActivitiesId) throws Exception {
+        
+        Session session = PersistenceManager.getCurrentSession();
+        Map<Long, Map<String, List<CompareOutput>>> activitiesCollection = new HashMap<Long, Map<String, List<CompareOutput>>>();
+        for (int i=0; i < ActivitiesId.length; i++) {
+            Map<String, List<CompareOutput>> ActivityList = compareActivities(ActivitiesId[i]);
```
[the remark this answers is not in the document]
MistreW: I have added an If condition to check for Null but it checks compareActivities(ActivitiesId[i]) method twice. I think this will degrade performance. shouldn't this check be on compareActivities(ActivitiesId[i]) method (AMPET-141)?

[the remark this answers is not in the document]
jdeanquin-dg: yes it will degrade performance what you need to do is return null in the method that compares to it previous version if it does not find an activity, and then don't add it to the map if the comparison is null

jdeanquin-dg:
The code looks good, please see the difference from @yosephtesfaye one pr #2661 I chose that but could have chosen this one too.

PR #2657, opened 2019-01-03
AMP-28752 Possibility to MOVE and RESIZE the legend on structures map in order to display all sectors

gmutuhu:
AMP-28752 Possibility to MOVE and RESIZE the legend on structures map in order to display all sectors


jdeanquin-dg:
replaced by #2707

PR #2682, opened 2019-01-08
AMP-28713 Update performance rule module FE after EP changes

gmutuhu:
AMP-28713 Update performance rule module FE after EP changes


mbaralo:
@vchihai please notice that the main PR has conflicts. Please along with validating this code, ensure it builds so it can be tested. Thanks

PR #2685, opened 2019-01-08
task/AMP-28122/activities api lists

nmandrescu:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/APIField.java
```diff
@@ -310,10 +321,16 @@ public void setDiscriminationConfigurer(
     public void setElementType(Class<?> elementType) {
         this.elementType = elementType;
     }
+    
+    @JsonIgnore
+    public boolean isSimpleItemType() {
+        return this.itemType != null && this.itemType != ActivityEPConstants.FIELD_TYPE_OBJECT;
```
developster: This might not work as expected. Please use equals method to compare strings.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/APIField.java
```diff
@@ -24,6 +24,9 @@
 
     @JsonProperty(ActivityEPConstants.FIELD_TYPE)
     private String fieldType;
+    
+    @JsonProperty(ActivityEPConstants.ITEM_TYPE)
+    private String itemType;
```
developster: I think now we have a good opportunity to merge fieldType, itemType and elementType in its own class that would define the type of an APIField. Maybe call it APIType? This way we can also better control how it is instantiated, for example that itemType and elementType is always null in all cases except lists.

Also what if we switch now from String to an Enum for itemType and fieldType field? It would help documentation.

amp/src/test/java/org/digijava/kernel/ampapi/endpoints/activity/validators/RequiredValidatorTest.java
```diff
@@ -193,9 +192,9 @@ private APIField fd(String required) {
 
     private APIField fdList(String required) {
         APIField fd = fd(FIELD_ALWAYS_REQUIRED);
-        fd.setElementType(String.class);
-        fd.setFieldType(FIELD_TYPE_LIST);
-        fd.setItemType(FIELD_TYPE_STRING);
+        fd.getApiType().setElementType(String.class);
+        fd.getApiType().setFieldType(FieldType.LIST);
+        fd.getApiType().setItemType(FieldType.LONG);
```
developster: How it is possible to have element type String but item type LONG? We should avoid such cases by deriving item type from element type is possible.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/field/APIType.java
```diff
@@ -0,0 +1,66 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.digijava.kernel.ampapi.endpoints.activity.ActivityEPConstants;
+
+/**
+ * @author Nadejda Manndrescu
+ */
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class APIType {
```
developster: It would be nice if this class was immutable. And constructor or static methods to construct these objects to do some checks. For example disallow to instantiate lists without specifying item/element type.

Also it appears that fieldType and itemType can be always derived from type and elementType.

amp/src/test/java/org/digijava/kernel/ampapi/endpoints/activity/validators/RequiredValidatorTest.java
nmandrescu: This is preexisting issue. I can refactor to avoid it.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/field/APIType.java
nmandrescu: ok

amp/src/test/java/org/digijava/kernel/ampapi/endpoints/activity/validators/RequiredValidatorTest.java
developster: I can see that item type has changed from string to long in 00d58ce43309cac8211c21b9924e2838017210a6. Or maybe I am misunderstanding the situation.

nmandrescu: There were no strong correlation between item and element type to make a test to fail. The `APIType` now won't allow setting item type explicitly, so it will always derive from element type.

PR #2700, opened 2019-01-11
AMP-21049: Cleanup activity import

developster:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityImporter.java
```diff
@@ -759,32 +655,6 @@ protected void updateRoleFundings() {
...
-                            break;
-                        }
-                }
-            
-                if (!found) {
-                    AmpOrgRole role = new AmpOrgRole();
-                    role.setOrganisation(f.getAmpDonorOrgId());
-                    role.setActivity(newActivity);
-                    role.setRole(f.getSourceRole());
-                    newActivity.getOrgrole().add(role);
```
nmandrescu: New org role generation doesn't seem to be covered through `InterchangeableBackReference`. Only existing org roles activity will be set.

developster: This code is creating any missing org roles that were specified in funding. However at the same time, we have a validator which does not allow activities with funding for which there is no corresponding org role defined. Thus this code is never invoked and unnecessary.

nmandrescu: Thanks for clarification, so when validator was added this code became obsolete - then I agree is safe to remove

PR #2719, opened 2019-01-16
Task/AMP-28803/adjust patch

jdeanquin-dg:


jdeanquin-dg:
View changed and removed so this pr is no longer needed

PR #2740, opened 2019-01-22
AMP-28824: Fixed headers for map tiles http response (3.3.2)

developster:


amp/WEB-INF/src/org/digijava/module/contentrepository/helper/NodeWrapper.java
```diff
@@ -644,7 +644,22 @@ public Calendar getCalendarDate() {
         }
         return null;
     }
-    
+
+    /**
+     * Return file size for a document or null if file size is unknown.
+     */
+    public Long getFileSize() {
+        Property fileSize = DocumentManagerUtil.getPropertyFromNode(node, CrConstants.PROPERTY_FILE_SIZE);
```
jdeanquin-dg: the code is similar to the one below, do you think we can unify it to avoid dupe?

PR #2784, opened 2019-02-08
AMP-28845: Allow saving reports with undefined option

vchihai:


developster:
We need to fix how filters are persisted. This PR just silently ignores 'undefined' values which is not better than throwing exception. Closing in favor of PR that fixes filters persistence.

PR #2800, opened 2019-02-14
AMP-20846: export tests

developster:


amp/src/test/java/org/digijava/kernel/ampapi/endpoints/activity/ActivityExporterTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+        AmpActivityVersion activity = new AmpActivityVersion();
+        activity.setActPrograms(programs);
+
+        Map<String, Object> jsonActivity = createExporter().export(activity).any();
+
+        assertThat(jsonActivity,
+                (Matcher) allOf(
+                        hasEntry(equalTo("primary_programs"),
+                                contains(actProgram(2L, 100F))),
```
nmandrescu: Since the id is defined in reports tests and if for any reason the id is changed, should it be safer to use theam.getId?

PR #2804, opened 2019-02-18
AMP-21055: Refactor IntcherchangeUtils

developster:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/FieldMap.java
```diff
@@ -0,0 +1,94 @@
+package org.digijava.kernel.ampapi.endpoints.activity;
```
nmandrescu: How about moving this class under `org.digijava.kernel.ampapi.endpoints.activity.field` or to `...common.field`?

PR #2816, opened 2019-02-22
AMPET-109: Show changes from all activities in audit logger

Wessi:
Add the functionality to export a full list of activities, compared to its previous versions.

### DIFF
--- a/amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
+++ b/amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
@@ -1,11 +1,8 @@
 package org.digijava.module.aim.action;
-import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Calendar;
-import java.util.Collection;
 import java.util.Date;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
@@ -13,8 +10,6 @@ import java.util.Map;
 import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import com.google.common.collect.ArrayListMultimap;
-import com.google.common.collect.Multimap;
 import org.apache.log4j.Logger;
 import org.apache.struts.action.ActionErrors;
 import org.apache.struts.action.ActionForm;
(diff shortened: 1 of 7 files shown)


amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
```diff
@@ -352,10 +353,38 @@ private void setAdvancemode(CompareActivityVersionsForm vForm, HttpServletReques
...
 
         return mapping.findForward("forward");
     }
 
+    public ActionForward compareAll(ActionMapping mapping, ActionForm form, HttpServletRequest request,
+                                    HttpServletResponse response) throws Exception {
+
+        CompareActivityVersionsForm vForm = (CompareActivityVersionsForm) form;
+
+        // We may have duplicate keys.. so time to switch to something like Multimap
```
jdeanquin-dg: If we have duplicate keys, we need to solve that at query level to avoid getting more results than needed. if we use a multi map,then we add overhead to the program (its slow and needs to be optimized, but we add a burden to remove duplicate objects). If the criteria returns dupes, please work it out to avoid them

amp/repository/aim/view/viewActivityDifferences.jsp
```diff
@@ -45,8 +45,17 @@
 	<html:hidden property="showMergeColumn" styleId="showMergeColumn"/>
 	<html:hidden property="method" styleId="method"/>
```
jdeanquin-dg: lets add a line break in the view so it is clear its a new activity

amp/repository/aim/view/viewActivityDifferences.jsp
```diff
@@ -287,6 +339,11 @@ if(document.getElementById('method').value == "enableMerge") {
     document.getElementById('saveButton').disabled = "disabled";
     document.getElementById('saveButton').style.display = 'none';
     $('#backButton').prop('value', '<digi:trn>Back to Audit Logger</digi:trn>');
+}else if (document.aimCompareActivityVersionsForm.method.value === "compareAll"){
+    document.getElementById('saveButton').disabled = "disabled";
+    document.getElementById('saveButton').style.display = 'none';
+    $('#MyTabs').replaceWith('<font color="red" size="2%"><strong><digi:trn>&nbsp;&nbsp;List of Activities Compared to their Previous Versions</digi:trn></strong></font>');
```
jdeanquin-dg: I wouldn't add this with JavaScript lets add it to the jsp code so its clear to be read

jdeanquin-dg: also as I added in my first comment, it would be easier and quicker to add the split in the table (an empty tr) would do

amp/repository/aim/view/viewActivityDifferences.jsp
jdeanquin-dg: I pasted a screenshot where the linebreak should go in order to give a bit of space between activities
image

amp/repository/aim/view/viewActivityDifferences.jsp
```diff
@@ -90,7 +99,50 @@
 	            		</div>
 	        		</td>
 				</tr>
-				
+
+					<%-- Iterate through the list of output collections for compareAll method... --%>
```
jdeanquin-dg: also try to reuse the code, not to duplicate the code that already shows the differences

PR #2821, opened 2019-02-25
Fix/amp 28671/no children select

ginchauspe:


ginchauspe:
This PR shows lots of changes in .java files from commits I didnt do, we need to investigate why.

PR #2830, opened 2019-03-01
Task/AMPET-109/Audit_Logger_List

MistreW:


amp/repository/aim/view/viewActivityDifferences.jsp
```diff
@@ -141,7 +141,69 @@
 						</logic:iterate>
 
 					</td></tr>
-				</logic:iterate>	
+				</logic:iterate>
```
jdeanquin-dg: same as wessi we should reuse as an include the jsp part that iterates over a single activity so we don't have duplicate code. Please refactor it so we include the duplicate code.and this way its easier to maintain

amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
```diff
@@ -357,5 +359,18 @@ public ActionForward viewDifferences(ActionMapping mapping, ActionForm form, Htt
 
         return mapping.findForward("forward");
     }
+    
+    public ActionForward viewAllDifferences(ActionMapping mapping, ActionForm form, HttpServletRequest request, HttpServletResponse response) throws Exception {
+        CompareActivityVersionsForm vForm = (CompareActivityVersionsForm) form;
+        Map<Long, Map<String, List<CompareOutput>>> listoutputCollectionGrouped = new HashMap<Long, Map<String, List<CompareOutput>>>();
```
jdeanquin-dg: this query should go to audit logger util, see @Wessi pr

amp/WEB-INF/src/org/digijava/module/aim/action/CompareActivityVersions.java
```diff
@@ -357,5 +359,12 @@ public ActionForward viewDifferences(ActionMapping mapping, ActionForm form, Htt
 
         return mapping.findForward("forward");
     }
-
+    
+    public ActionForward viewAllDifferences(ActionMapping mapping, ActionForm form, HttpServletRequest request, HttpServletResponse response) throws Exception {
+        CompareActivityVersionsForm vForm = (CompareActivityVersionsForm) form;
+        Map<Long, Map<String, List<CompareOutput>>> listoutputCollectionGrouped = new HashMap<Long, Map<String, List<CompareOutput>>>();
+         listoutputCollectionGrouped = ActivityVersionUtil.compareActivities(AuditLoggerUtil.getListOfActivitiesFromAuditLogger());        
+         vForm.setListoutputCollectionGrouped(listoutputCollectionGrouped);
```
jdeanquin-dg: vForm.setListoutputCollectionGrouped is populating the internal variable listoutputCollectionGrouped and then the jsp file is using   return this.outputCollectionGrouped.entrySet(); which is null. also the return type of the entryset is not consistent. I think you have might have code missing to commit since it always returns nullpointer and Im sure for you worked

PR #2832, opened 2019-03-01
AMP-28898: Fix pledge programs filters

vchihai:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/filters/PledgesProgramFilterListManager.java
```diff
@@ -83,6 +87,19 @@ private PledgesProgramFilterListManager() {
         return filterIds;
     }
     
+    @Override
+    protected List<AmpActivityProgramSettings> getProgramSettings() {
```
jdeanquin-dg: Hey I see that this method is very similar than the one we have in ProgramFilterListManager# getProgramSettings the only difference I see is where we are getting the view definitions from. Can we try to abstract to avoid having duplicate code?
Thanks

vchihai: Yes Julian, of course.

PR #2842, opened 2019-03-06
AMP-28975: Backport contentmanager and lucene changes

vchihai:


jdeanquin-dg:
the lucene version was not the problem

PR #2862, opened 2019-03-14
Fix/amp 28885/keep selection

ginchauspe:


ginchauspe:
Time test keeps failing for no reason, will have to recreate the PR.

PR #2874, opened 2019-03-27
AMP-29021: Business rules validator should be used after validating the format of subelements

vchihai:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/validators/TreeCollectionValidator.java
```diff
@@ -147,7 +147,7 @@ private boolean isPresentParentProgramInCollection(Long programId, Set<Long> idV
                 } 
             }
         } catch (DgException e) {
-            throw new RuntimeException(e);
+            logger.error(e.getMessage(), e);
```
nmandrescu: In fact if we expect "valid" input here, then throwing a `RuntimeException` is right to flag that something is wrong (like in this case we had to run business rule check after subelements validation). Please restore.

vchihai: I've kept logging the message. It was hard to understand the root of the problem without stack trace.

nmandrescu: This is fine, logging more doesn't hurt :) but good that you restored the `RuntimeException`.

PR #2885, opened 2019-04-01
task/AMP-28971/generalize some api fields

nmandrescu:


amp/src/test/java/org/digijava/kernel/ampapi/endpoints/activity/validators/ValueValidatorTest.java
```diff
@@ -117,4 +127,19 @@ public void testPossibleValueFYInvalid() throws Exception {
                 valueValidator.isValid(importer, newFieldParent, fyFieldDescription, FY_FIELD));
         assertEquals(ActivityErrors.FIELD_INVALID_VALUE, valueValidator.getErrorMessage());
     }
+
+    @Test
+    public void testDraftSubmissionDraftFMDisabled() throws Exception {
```
vchihai: The exception is never thrown. I think we need to remove all the Exception throws (including above tests)

nmandrescu: Oh, you are right, I took an existing test case from above as a template and didn't notice this, let me remove them.

PR #2891, opened 2019-04-05
task/AMP-29039/ignore unkown fields

nmandrescu:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityInterchangeUtils.java
```diff
@@ -71,11 +72,11 @@ public static JsonBean importActivity(JsonBean newJson, boolean update, Activity
         ActivityImporter importer = new ActivityImporter(activityFields, rules);
         List<ApiErrorMessage> errors = importer.importOrUpdate(newJson, update, endpointContextPath);
 
-        return getImportResult(importer.getNewActivity(), importer.getNewJson(), errors);
+        return getImportResult(importer.getNewActivity(), importer.getNewJson(), errors, importer.getWarnings());
```
vchihai: Why we don't pass the importer as a parameter? 
GetImportResult is called only from this line of code.

vchihai: The errors also are part of the impoter.

nmandrescu: If you wanted to say this way that it would be good to refactor the method input params, then I would like to let you know that I thought about refactoring this method input as well. However exactly for the reason that it is used only here, it seemed simpler to define new local variables through input method params. Therefore I decided it is not that much helpful to refactor it.

vchihai: Ok, let's create another issue to refactor the way how importer result is returned.

PR #2894, opened 2019-04-08
AMP-29031: Fix saving activity creator

vchihai:


amp/WEB-INF/src/org/dgfoundation/amp/onepager/util/ActivityUtil.java
```diff
@@ -345,7 +342,11 @@ public static void updateModifyCreateInfo(AmpActivityVersion activity, AmpTeamMe
     
         activity.setModifiedBy(modifier);
     
-        if (activity.getAmpActivityId() == null || activity.getActivityCreator() == null) {
+        if (activity.getCreatedDate() == null) {
+            activity.setCreatedDate(updateDate);
+        }
+    
+        if (activity.getActivityCreator() == null) {
```
nmandrescu: I was wondering that in case there is/was any bug and creator was not set when an activity was created, then it shouldn't be set during update. I checked for how creator was set sometime ago and indeed it was configured only when activity id was null: https://github.com/devgateway/amp/pull/2569/files#diff-e60b89bba56d1d3c749d575470a1acc0L203

Please set creator only when activity is newly created.

PR #2906, opened 2019-04-10
AMP-28993: Handle missing fields during the update of activity in api

vchihai:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ObjectImporter.java
```diff
@@ -194,13 +196,14 @@ private boolean validateAndImport(Object newParent, APIField fieldDef,
             Map<String, Object> newJsonParent, String fieldPath) {
         String fieldName = getFieldName(fieldDef, newJsonParent);
         String currentFieldPath = (fieldPath == null ? "" : fieldPath + "~") + fieldName;
+        
         Object newJsonValue = newJsonParent == null ? null : newJsonParent.get(fieldName);
 
-        boolean isValidFormat = formatValidator.isValid(this, newJsonParent, fieldDef, currentFieldPath, errors);
+        boolean isValidFormat = formatValidator.isValid(this, newParent, newJsonParent, fieldDef, currentFieldPath);
         if (isValidFormat) {
             isValidFormat = validateSubElements(fieldDef, newParent, newJsonValue, currentFieldPath);
```
developster: Please check if validateSubElements will modify newParent. Looks like fields that are types or collections will be set to null in case they are missing in json.

PR #2909, opened 2019-04-11
AMP-29069: Fix Lucene indexing for concurrent actions

vchihai:


amp/WEB-INF/src/org/digijava/module/aim/util/LuceneUtil.java
```diff
@@ -741,60 +743,34 @@ public Integer execute(Connection conn) throws SQLException {
...
-        IndexReader indexReader;
+        IndexWriter indexWriter = null;
         try {
-            indexReader = IndexReader.open(FSDirectory.open(new File(idx)), false);
-            //listDocuments(indexReader);
-            int ret = indexReader.deleteDocuments(term);
-            indexReader.close();
-            return ret;
+            IndexWriterConfig indexWriterConfig = new IndexWriterConfig(Version.LUCENE_36, LuceneUtil.ANALYZER);
+            Directory directory = FSDirectory.open(new File(idx));
```
jdeanquin-dg: we need to set LOCK_OBTAIN_WAIT_FOREVER here too. If you post a lot of updates in parallel you get the cannot acquire lock exception. I tried with 5 and I got the exception

PR #2920, opened 2019-04-16
Fix/AMP-29070/js overload

ginchauspe:


amp/TEMPLATE/ampTemplate/node_modules/amp-filter/src/models/generic-filter-model.js
```diff
@@ -78,9 +74,9 @@ module.exports = BaseFilterModel.extend({
 
     reset: function () {
         var tree = this.get('tree');
-        if (tree) {
+        if (tree && tree.get('numSelected') > 0) {
             // Locations need a special treatment because they can have parents selected without their children.
-            if (tree.get('name') === Constants.LOCATIONS) {
+            if (tree._isLocation() && tree.get('id') === -1 && tree.get('parent') === undefined) {
```
jdeanquin-dg: check if its needed to ask if id == -1 and parent undefined. If so, then the else should check the same thing when the model is not a location

ginchauspe: done.

PR #2941, opened 2019-04-23
AMP-29052: Update correct the version activity when draft is changed

vchihai:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityImporter.java
```diff
@@ -234,9 +236,12 @@ private void init(JsonBean newJson, boolean update, String endpointContextPath)
             validateAndImport(newActivity, newJsonParent);
             if (errors.isEmpty()) {
                 prepareToSave();
-
+                
+                boolean newActitityDraft = Boolean.TRUE.equals(newActivity.getDraft());
+                Boolean draftChange = oldActivityDraft != newActitityDraft;
```
nmandrescu: I was looking for an explicit method that will detect draft change since it should be part of `ActivityUtil` due to its internal logic nature (e.g. the default setup for `oldActivityDraft = false`):
1. API would call that `detectDraftChange(oldActivity, newActivityDraft)` to pass to `draftChange` param of `saveActivityNewVersion`-v2.
2. The rest callers would call `saveActivityNewVersion`-v1 as before, however this internally would pass to `saveActivityNewVersion`-v2 as `detectDraftChange(a, draft)`

PR #2942, opened 2019-04-24
AMP-29126: Refactor the mechanism of rejecting the activity

vchihai:


vchihai:
I've checked the body of the method and it is correct.
ActivityUtil.setActivityStatus() checks if the activity will be in the draft state (not the draft state before saving) and set the status corresponding to the value. 
It can be read like this: if the activity will be saved as draft, set the status -> rejected/started/edited

It is different from checking if an activity can be rejected or not. The method setActivityStatus() is called after the activity was set to draft.

PR #2945, opened 2019-04-26
Fix/amp 29012/agencies levels

ginchauspe:


jdeanquin-dg:
Please before merging check the comment in jira

PR #2948, opened 2019-04-29
AMP-29005: Standardize fields access during data importing in API

vchihai:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityImporter.java
```diff
@@ -23,7 +22,6 @@
 import org.dgfoundation.amp.ar.AmpARFilter;
 import org.dgfoundation.amp.newreports.AmountsUnits;
 import org.dgfoundation.amp.onepager.util.ActivityGatekeeper;
-import org.dgfoundation.amp.onepager.util.ActivityUtil;
```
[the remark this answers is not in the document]
vchihai: agree, I thought it was done in another issue.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ActivityImporter.java
```diff
@@ -237,9 +235,9 @@ private void init(JsonBean newJson, boolean update, String endpointContextPath)
             validateAndImport(newActivity, newJsonParent);
             if (errors.isEmpty()) {
                 prepareToSave();
-                boolean draftChange = ActivityUtil.detectDraftChange(newActivity, oldActivityDraft);
```
nmandrescu: I am afraid this will be a regression: before `draftChange` for API was calculated based on `oldActivityDraft`. Please restore.

vchihai: Indeed, you are right. I still get confused about why this regression appeared.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/field/APIField.java
```diff
@@ -32,6 +34,9 @@
 
     @JsonUnwrapped
     private APIType apiType;
+    
+    @JsonIgnore
+    private Class<?> internalFieldType;
```
developster: After an analysis with Viorel, we've come up with the conclusion that this field is not required. Also in APIType, elementType is no longer needed. If APIType#fieldType is anything but list then type is the class of the field. If APIType#fieldType is list, then type is the class of the elements of that list.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/field/APIField.java
```diff
@@ -102,6 +107,12 @@
 
     @JsonProperty(ActivityEPConstants.COMMON_POSSIBLE_VALUES)
     private String commonPossibleValuesPath;
+    
+    @JsonIgnore
+    private TranslationSettings.TranslationType translationType;
+
+    @JsonIgnore
+    private boolean isCollection;
```
developster: This is probably needed for cases like ppc_amount where the field is an object in json but object model it is a discriminated collection with just one element. I think this creates unnecessarily complex code in ObjectImporter/Exporter. What if DiscriminatedFieldAccessor returns just a single object for ppc_amount instead of a list with a single element? In that case this field will not be required at all.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/ObjectImporter.java
```diff
@@ -217,107 +215,66 @@ private boolean validateAndImport(Object newParent, APIField fieldDef,
...
-
-                if (col == null) {
-                    col = (Collection) valueConverter.getNewInstance(parentObj, field);
-                }
-
-                if (idOnly && jsonValue != null && !fieldDef.getApiType().isSimpleItemType()) {
-                    Object res = valueConverter.getObjectById(fieldDef.getApiType().getType(), jsonValue);
-                    col.clear();
-                    col.add(res);
+                Collection collection = (Collection) apiField.getFieldAccessor().get(parentObj);
```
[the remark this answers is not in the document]
vchihai: Agree with the 2nd statement, that we need to create a copy of collection when the value of the field is fetched using SimpleFieldAccessor

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/field/APIField.java
vchihai: After investigating the code, it would be difficult at this moment to change the type of DiscriminatedFieldAccessor  without full refactoring of Object Importer. It will be done in a separate ticket.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/field/APIField.java
vchihai: Agree

PR #2971, opened 2019-05-24
 AMP-29148: Convert eastern arabic numbers for activities list in GIS

vchihai:


jdeanquin-dg:
There are some fixes to be done front end side, merging this ticket so we keep only one open

PR #2974, opened 2019-05-28
AMP-29190: Log email recipients for demo

vchihai:


jdeanquin-dg:
Ticket already tested, merging

PR #2983, opened 2019-06-03
AMP-29213: Unique Validator

developster:


nmandrescu:
I interrupted the review, since I saw the ticket is in progress and I see it is still in progress. So I am flagging as postponed. Please remove the label when ready for review.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/field/FieldsEnumerator.java
```diff
@@ -169,6 +171,15 @@ protected APIField describeField(Field field, FEContext context) {
             }
         }
 
+        List<ConstraintDescriptor> fieldConstraintDescriptors = new ArrayList<>();
+
+        if (apiField.getUniqueConstraint() != null) {
+            Map<String, String> args = ImmutableMap.of("field", apiField.getUniqueConstraint());
+            Set<Class<?>> groups = ImmutableSet.of();
+            fieldConstraintDescriptors.add(new ConstraintDescriptor(
+                    UniqueValidator.class, args, groups, ConstraintDescriptor.ConstraintTarget.FIELD));
```
nmandrescu: Looks like you picked the option 2 from approach so that some validators are managed through `FieldsEnumerator` as well.

developster: This is required in order to avoid breaking API clients. In API unique validator is enabled when a tree validator is present. If we remove that requirement then it would be easy to move the validator description into `@InterchangeableValidator` annotation.

nmandrescu: Thanks for clarifying, though I also saw the underlying reasons of simplicity (and I suppose you meant `not` easy to move).

PR #2987, opened 2019-06-05
AMP-29201: Implement RTL and numbers in Global Settings admin page

vchihai:


vchihai:
The values are not updated. Only option label is converted.

PR #2988, opened 2019-06-06
AMP-29202: Implement RTL in Visibility Manager

vchihai:


vchihai:
Let's create a separate ticket for this.

PR #2989, opened 2019-06-06
AMP-28993: Handle missing fields during the update of activity in api

vchihai:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/validators/UUIDValidator.java
```diff
@@ -22,13 +22,24 @@
...
+    }
+
+    @Override
+    public boolean isValid(ObjectImporter importer, Object currentObject, Map<String, Object> newFieldParent,
             APIField fieldDescription, String fieldPath) {
         
         if (fieldDescription.getFieldName().equals(FieldMap.underscorify(ActivityFieldsConstants.UUID))) {
-            String uuid = StringUtils.trim((String) newFieldParent.get(fieldDescription.getFieldName()));
+            String uuid = null;
+            if (!newFieldParent.containsKey(fieldDescription.getFieldName()) && importer.isUpdate()) {
```
developster: Looks like it doesn't matter if we update or not. While we don't have this situation but even during insert a field could have some default value that is not null.

nmandrescu:
@vchihai, please fix conflicts, @developster please prioritize testing afterwards, thanks

vchihai:
The conflicts were fixed

PR #2990, opened 2019-06-07
AMP-29214: Regex validator

developster:


amp/WEB-INF/src/org/digijava/module/aim/dbentity/AmpContactEmailProperty.java
```diff
@@ -3,15 +3,19 @@
...
 /**
  * @author Octavian Ciubotaru
  */
 public class AmpContactEmailProperty extends AmpContactProperty {
 
     @Interchangeable(fieldTitle = "Value", required = ALWAYS, importable = true,
-            regexPattern = ActivityEPConstants.REGEX_PATTERN_EMAIL)
+            interValidators = @InterchangeableValidator(
+                    value = RegexValidator.class,
+                    attributes = "regex=" + ActivityEPConstants.REGEX_PATTERN_EMAIL))
```
nmandrescu: Would be also good to use a constant for "regex" here and in validator

PR #3011, opened 2019-06-17
AMP-29197: Remove JsonBean from ResourceManager

vchihai:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/filetype/ResourceManager.java
```diff
@@ -84,19 +82,17 @@
                     + "    }\n"
                     + "}\n"
                     + "```\n")
```
nmandrescu: Please move the above description at `ResourceManagerSettings`. Also the settings are small in number, so we can use concrete fields instead of a generic map, which allows a better documentation and description.

PR #3020, opened 2019-06-20
AMP-29210: Unique activity title validator

developster:


amp/WEB-INF/src/org/dgfoundation/amp/ar/viewfetcher/SQLUtils.java
```diff
@@ -320,8 +320,8 @@ public String toSqlString(Criteria criteria, CriteriaQuery criteriaQuery) throws
                     throw new HibernateException("We do not support multiple identifiers just yet!");
 
                 if ( dialect instanceof PostgreSQLDialect ) {
-                    //AMP-15628 - the replace of "this_." with "" inside the ids and columns was removed
-                    String ret=" "+ids[0]+" = any(contentmatch('"+entityName+"','"+columns[0]+"','"+locale+"', ?)) OR ";
+                    String ret = " " + ids[0] + " = any(contentmatch('" + entityName + "','" + propertyName + "','"
```
nmandrescu: Was there any issue that needed to replace the matching `columns[0]` with `propertyName`? If this is only because so far all `propertyName` match `column` name, then let's better leave it as `columns[0]` for future.

developster: There are two issues. First, column name includes alias. Second is that column name could be different from property name, for example donorObjective vs donor_objective.

Changed when I was using old (now deleted) method to find duplicate titles.

nmandrescu: The method is `toSqlString`, so it should use `donor_objective` (as original) instead of `donorObjective` (this change). If there are issues with alias - please add a comment, we can look into this separately.

amp/WEB-INF/src/org/dgfoundation/amp/ar/viewfetcher/SQLUtils.java
```diff
@@ -320,8 +320,8 @@ public String toSqlString(Criteria criteria, CriteriaQuery criteriaQuery) throws
...
+                            + locale + "', ?)) OR ";
                     ret+=" unaccent(" + columns[0] + ") ilike " +  "unaccent(?)";
```
nmandrescu: Also this line still uses `columns[0]`

amp/WEB-INF/src/org/dgfoundation/amp/ar/viewfetcher/SQLUtils.java
nmandrescu: Sorry, you are right, since is used for amp_content_translation

PR #3024, opened 2019-06-20
AMP-29278: Refactor common endpoint to retrieve FM values. Remove JsonBean reference

vchihai:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/fm/FMService.java
```diff
@@ -22,49 +22,49 @@
...
+        return fmSettings;
     }
     
-    private static void provideModulesDetails(Map<String, Object> fmSettingsResult, FMSettingsConfig config) {
+    private static void provideModulesDetails(FMSettings fmSettings, FMSettingsConfig config) {
         List<String> detailModules = config.getDetailModules();
-        if (detailModules == null || detailModules.size() == 0) {
+        
+        FMSettingsTree settingsTree = new FMSettingsTree();
+        FMSettingsFlat settingsFlat = new FMSettingsFlat();
```
nmandrescu: I think it would be nicer to initialize them within `fmSettings` object and then below use `fmSettings.getFmSettings()..getModules().put(...`

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/fm/FMSettings.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.annotations.ApiModelProperty;
+import org.digijava.kernel.ampapi.endpoints.common.EPConstants;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public class FMSettings<T> {
```
nmandrescu: Let's use a different naming, may be `FMResult`? Since it has inside another `fmSettings` property which is confusing.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/fm/FMSettingsFlat.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Set;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import io.swagger.annotations.ApiModelProperty;
+
+public class FMSettingsFlat {
+    
+    @ApiModelProperty(value = "list of FM settings",
+            example = "### Use cases\n"
+                    + "1. Detail flat & fully enabled paths => /Activity Form/Organiation/Donor Organization\n"
```
nmandrescu: Since this is for flat EP branch only, we no longer need to clarify the `Detail flat` text. Or you can move up to the value `list of FM settings in flat format`

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/fm/FMSettingsTree.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+import io.swagger.annotations.ApiModelProperty;
+
+public class FMSettingsTree {
+    
+    @ApiModelProperty(value = "list of FM settings listed as a tree",
+            example = "Detail tree & fully enabled paths =>\n"
```
nmandrescu: Here as well we can avoid repeating `detail tree`

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/fm/FMSettingsTree.java
```diff
@@ -0,0 +1,41 @@
...
+                    + "    \"REPORTING\": {\n"
+                    + "         \"__enabled\" : true, // omitted if fullEnabledPaths are requested (same below)\n"
+                    + "         \"Measures\": {\n"
+                    + "             \"__enabled\" : true,\n"
+                    + "             \"Actual Disbursements\": {\n"
+                    + "                 \"__enabled\" : true\n"
+                    + "             },\n"
+                    + "             ...\n"
+                    + "    ```\n")
+    private Map<String, Object> modules = new HashMap<>();
```
nmandrescu: Looking at structure I think you can do this:
```
public class FMSettingsTree {
   @JsonProperty("__enabled")
    boolean enabled;
   
   private Map<String, FMSettingsTree> tree; // still use JsonAnyGetter for it
}
```

nmandrescu: And use `@JsonUnwrapp` for `T fmSettings`;

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/fm/FMService.java
vchihai: We don't know the type of T during initialization. 
Also, it means to create an interface with modules field as a generic Map. (the modules are different map in the classes)

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/fm/FMSettings.java
vchihai: agree

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/fm/FMService.java
nmandrescu: For parametereized class in constructor:
```
Class<T> clazz = (Class<T>) ((ParameterizedType) this.getClass().getGenericSuperclass()).getActualTypeArguments()[0];
```
And for modules we can cast when accessing fmSettings.

Anyway, was just a thought, and of course we can leave as is.

vchihai: It would be better to leave as it is ;)

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/common/fm/FMSettingsTree.java
vchihai: Great idea!
P.S.: we don't need to use @JsonUnwrapp

nmandrescu: True, no need for unwrapping, I forgot again we report "fm-settings" explicitly.

amp/WEB-INF/src/org/dgfoundation/amp/visibility/data/FMTree.java
```diff
@@ -89,20 +89,23 @@ public FMTree(Map<String, FMTree> entries, boolean enabled) {
...
      */
-    public Map<String, Object> asMap(boolean fullEnabledPaths) {
-        Map<String, Object> mapTree = new HashMap<>();
-        if (!fullEnabledPaths || this.enabled)
+    public FMSettingsTree asFmSettingsTree(boolean fullEnabledPaths) {
+        FMSettingsTree fmSettingsTree = new FMSettingsTree();
+        if (!fullEnabledPaths || this.enabled) {
             if (!fullEnabledPaths) {
-                mapTree.put(EPConstants.FM_ENABLED, this.enabled);
+                fmSettingsTree.setEnabled(this.enabled);
```
nmandrescu: Oh, I have just noticed that the existing `FMTree` matches `FMSettingsTree` (funny to notice the consistency over time :D), so may be just to use it directly instead?

PR #3090, opened 2019-07-26
AMP-29264: The approval date should be the same in allowed approved constraint test

vchihai:


amp/src/test/java/org/digijava/module/aim/validator/AllowedApproverConstraintTest.java
```diff
@@ -177,11 +177,13 @@ public void testInvalidApprovedByForValidatedActivityWhenNotMatchingModifiedByAn
     }
 
     @Test
-    public void testInvalidApprovedByForUnvalidatedActivityWhenNotMatchingModifiedByAndPastAppDateValidateNewOnly() {
+    public void testInvalidApprovedByForUnvalidatedActivityWhenNotMatchingModifiedByAndPastAppDateValidateNewOnly() throws InterruptedException {
```
jdeanquin-dg: this exception is never thrown

PR #3141, opened 2019-09-27
AMPUI-80 call activity endpoint. load fields definition, load possibl…

jdeanquin-dg:
…e values, load FM tree, refactor common possiblevalues to ampui


amp/TEMPLATE/reamp/modules/activity/preview/containers/App.jsx
```diff
@@ -54,31 +51,34 @@ class App extends Component {
...
         }
     }
 
     render() {
-        return (
-            <ActivityPreviewUI activity={activity} activityContext={activityContext} />
-        );
+        console.log(this.props);
+        if (this.props.activityReducer.isActivityLoading) {
+            return (<div> LOADING ACTIVITY </div>)
```
ginchauspe: In a future phase we will add translate here.

amp/TEMPLATE/reamp/modules/activity/preview/api/ActivityApi.jsx
```diff
@@ -0,0 +1,48 @@
+import {fetchJson, postJson} from 'amp/tools/index';
+import {ACTIVITY_API, FIELDS_DEFINITION_API, POSSIBLE_VALUES_API, FM_API} from '../common/Constants.jsx'
+
+
+export default class ActivityApi {
+    static getActivity(activityId) {
+        return new Promise((resolve, reject) => {
+            fetchJson(ACTIVITY_API + activityId).then((activity) => {
```
ginchauspe: my ES6 is a bit rusty, could you confirm if in these cases is not necessary to put a return before beforeJson?

amp/TEMPLATE/reamp/modules/activity/preview/containers/App.jsx
jdeanquin-dg: yes once we implement the translate module we will. Good catch

amp/TEMPLATE/reamp/modules/activity/preview/api/ActivityApi.jsx
jdeanquin-dg: I think we do, let me check. Also we could enable eslint but only for the preview module

PR #3143, opened 2019-10-03
AMPUI-89 load activityContext from api

jdeanquin-dg:


jdeanquin-dg:
changes already validated by @ginchauspe  I just fixed checkstyle

PR #3201, opened 2019-12-18
AMPUI-97 Implement Logger Manager

jdeanquin-dg:


amp/TEMPLATE/reamp/modules/activity/preview/utils/LoggerManager.jsx
```diff
@@ -0,0 +1,31 @@
+import * as log from 'loglevel';
+
+export default class LoggerManager {
+
+    constructor(module) {
+        this.logger = log.getLogger(module);
+        if (process.env.NODE_ENV && process.env.NODE_ENV === 'production') {
```
ginchauspe: Can we have this string as a constant? (here and in other files).

PR #3224, opened 2020-02-27
AMP-29496: Make year of publication property importable in resource endpoint (3.5)

vchihai:


vchihai:
The branche was updated with future/3.5 changes.

PR #3290, opened 2020-04-15
Task/amp-29511/alert supported browser future

jdeanquin-dg:


ginchauspe:
Please check if our version of jquery has the same functionality.

jdeanquin-dg:
Francis tested confirmed its working

PR #3302, opened 2020-05-01
Task/AMP-29589/build process

jdeanquin-dg:
…up component to load translations


jdeanquin-dg:
code review will be done at the end since we still have a lot of intermediate code

PR #3305, opened 2020-05-07
AMP-29583 make donor country non ssc. move template from org to locat…

jdeanquin-dg:
…ion. Add extra info to location. enable countries without children and first level only

### DIFF
--- a/amp/WEB-INF/src/org/digijava/module/aim/util/LocationSkeleton.java
+++ b/amp/WEB-INF/src/org/digijava/module/aim/util/LocationSkeleton.java
@@ -14,15 +14,8 @@ public class LocationSkeleton extends HierEntitySkeleton<LocationSkeleton> {
     protected Double lat;
     protected Double lon;
 
-    public LocationSkeleton(LocationSkeletonBuilder locationSkeletonBuilder) {
-        super(locationSkeletonBuilder.getId(),
-                locationSkeletonBuilder.getLocName(),
-                locationSkeletonBuilder.getCode(),
-                locationSkeletonBuilder.getParentId());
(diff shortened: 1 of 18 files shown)


amp/WEB-INF/src/org/digijava/module/aim/util/LocationSkeletonBuilder.java
```diff
@@ -0,0 +1,88 @@
+package org.digijava.module.aim.util;
+
+public class LocationSkeletonBuilder {
+    private Long id;
```
vchihai, line 4: Usually, the builder should contain the insatnce of the target class and the method could be:
withId(Long id) {
    skeleton.setId(id);
}

and the instance can be get with build() method
ActivityBuilder could be used asa model.

amp/WEB-INF/src/org/digijava/module/aim/util/LocationSkeleton.java
```diff
@@ -6,38 +6,66 @@
...
 public class LocationSkeleton extends HierEntitySkeleton<LocationSkeleton> {
     protected Long cvId; // category value id
+    protected Long templateId;
+    protected Double lat;
+    protected Double lon;
 
-    public LocationSkeleton(Long id, String locName, String code, Long parentId, Long cvId) {
-        super(id, locName, code, parentId);
-        this.cvId = cvId;
+    public LocationSkeleton(LocationSkeletonBuilder locationSkeletonBuilder) {
```
vchihai, line 17: It is avoidable to fetch the build in the instance constructor.
The pattern is that the builder creates the instance.

PR #3306, opened 2020-05-08
AMP-29521: API, allow multiple values for pickIdOnly

jdeanquin-dg:


jdeanquin-dg:
Code committed by Octavian this pr replaces #3293

PR #3326, opened 2020-06-10
Task/amp 29642/fix imp loc

jdeanquin-dg:
The previous pr was for develop and the tables are not the same


mbaralo:
The patch was manually applied on the DB to unblock the testing. The code has been tested already. Merging.

PR #3348, opened 2020-07-06
AMP-29566: Machine Translation API (future 3.5)

vchihai:


vchihai:
We need to identify if the changes in develop are present in future too, the other one related to this issue

PR #3352, opened 2020-07-16
Task/amp-29684/xls export

jdeanquin-dg:


amp/TEMPLATE/reampv2/src/modules/sscdashboard/utils/exportUtils.js
```diff
@@ -0,0 +1,204 @@
...
+ * Code from VIFAA project.
+ */
+export const dataDownloadStyles = {
+    totalLabels: {
+        alignment: {
+            vertical: 'middle',
+            horizontal: 'middle'
+        },
+        font: {
+            name: 'Calibri',
```
ginchauspe, line 14: We could move some of the strings to constants if they are going to be used elsewhere.

PR #3365, opened 2020-07-31
AMP-29702 fix map config

jdeanquin-dg:


mbaralo:
tested and works. please merge

PR #3378, opened 2020-08-14
AMP-29720: Geo Coder process and status EP

developster:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/geocoding/GeoCoderEndpoint.java
```diff
@@ -0,0 +1,79 @@
...
+    @ApiOperation("Returns geo coding status")
+    @ApiResponses({
+            @ApiResponse(code = HttpServletResponse.SC_OK, message = "geo coding status", response = GeoCoding.class),
+            @ApiResponse(code = HttpServletResponse.SC_NOT_FOUND, message = "no geo coding in progress") })
+    @ApiMethod(id = "process", authTypes = AuthRule.IN_WORKSPACE)
+    @GET
+    @Path("status")
+    @Produces(MediaType.APPLICATION_JSON + ";charset=utf-8")
+    public GeoCoding getProcessingStatus() {
+        GeoCoding status = service.getProcessingStatus();
```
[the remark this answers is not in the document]
vchihai: Maybe it would be better to rename the ep path to /info?

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/geocoding/GeoCoderEndpoint.java
```diff
@@ -0,0 +1,79 @@
...
+        if (status == null) {
+            throw new WebApplicationException(
```
vchihai, line 72: We shouldn't throw an error if the no geocoding is in progress or is not available in current ws.
If the geocoding is not available for the current user, the status should be "NOT_AVAILABLE" with additional info about the user/ws
If no geocoding is present in the system, we could add a new status: "OPEN" or "AVAILABLE" for geocoding

amp/WEB-INF/src/org/digijava/kernel/geocoding/service/GeoCodingService.java
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        // TODO update locations only if changed
+        client.sendLocations(LocationUtil.getAllVisibleLocations());
+
+        List<AmpActivityVersion> activities = new ArrayList<>();
+        for (Long activityId : activityIds) {
+            try {
+                AmpActivityVersion v = ActivityUtil.loadActivity(activityId);
+                if (!v.getLocations().isEmpty()) {
```
vchihai, line 54: If the activity has a location at national level (country assigned), it shouldn't be an error.

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/geocoding/GeoCoderEndpoint.java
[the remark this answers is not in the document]
developster: Let's call it a process. Currently only one geo coding process can happen at any given time.
Get /process will return the current process.

amp/WEB-INF/src/org/digijava/kernel/geocoding/service/GeoCodingService.java
developster: Ok.

PR #3392, opened 2020-08-20
AMP-29740 first batch

jdeanquin-dg:


jdeanquin-dg:
Replaced by #3401 3401

PR #3394, opened 2020-08-27
Task/ampoffline 1528/category values

ginchauspe:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/AmpPossibleValuesDAO.java
```diff
@@ -10,27 +10,55 @@
...
+        List<String> prefixes = ((List<String>) TLSUtils.getRequest().getAttribute(PREFIXES));
+        if (prefixes == null) {
+            prefixes = ActivityUtil.loadWorkspacePrefixesIntoRequest();
+        }
+        if (prefixes != null && prefixes.size() > 0) {
+            queryString += " WHERE acv.ampCategoryClass.keyName IN (";
+            for (String prefix : prefixes) {
+                queryString += "'" + prefix + discriminatorOption + "', ";
+            }
+            queryString += "'" + discriminatorOption + "') ORDER BY acv.id";
```
jdeanquin-dg, line 51: If im not wrong if there is not prefix the query won't run du to the lack of where clause

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/activity/DiscriminatedFieldAccessor.java
```diff
@@ -43,18 +46,23 @@ public DiscriminatedFieldAccessor(FieldAccessor target, String discriminatorFiel
...
     public Object get(Object targetObject) {
         Collection collection = getWrappedCollection(targetObject);
+        String prefix = "" + TLSUtils.getRequest().getAttribute(WORKSPACE_PREFIX);
         if (multipleValues) {
             List<Object> filteredItems = new ArrayList<>();
             for (Object item : collection) {
-                if (getDiscriminationValue(item).equals(discriminatorValue)) {
+                // AMPOFFLINE-1528
+                if (getDiscriminationValue(item).equals(discriminatorValue)
+                        || getDiscriminationValue(item).equals(prefix + discriminatorValue)) {
```
jdeanquin-dg, line 55: I think we need to check with prefix and if not use without not both at a time as we do in online. but I might be wrong, lets discuss

ginchauspe: We need both.

PR #3422, opened 2020-09-30
Task/AMP-29787/adjust lvl 3

ginchauspe:


amp/WEB-INF/src/org/digijava/module/aim/util/ProgramUtil.java
```diff
@@ -148,17 +148,16 @@ public static AmpTheme getThemeByCode(String code) {
         }
 
 
-        public static AmpTheme getTheme(String name) {
+        public static AmpTheme getTheme(String id) {
```
jdeanquin-dg, line 151: you are changing the search variable and its being used in 
AddTheme.checkExisting#151

jdeanquin-dg (changes requested):

PR #3453, opened 2020-11-03
AMP-29843: Add default used value in category value for implementation location

vchihai:


jdeanquin-dg (changes requested):
fix tests

vchihai:
The test failed due to previous changes in the main branch. Have fixed the problematic test.

jdeanquin-dg (approved):

PR #3498, opened 2020-12-07
AMP-29885: Add mechanism of mapping SDG programs in API

vchihai:


jdeanquin-dg (changes requested):
Let's not tight to SGD only since SGD is a program, let's add a more generic name. We are mapping two programs and adding a restriction for the secondary((mapped)

jdeanquin-dg (approved):

PR #3524, opened 2020-12-22
Fix/amp 29923/no mapping

ginchauspe:


amp/TEMPLATE/reampv2/src/modules/ndddashboard/components/styles.css
```diff
@@ -36,7 +36,7 @@
 }
 
 .loading {
-    background-image: url("/assets/loading.gif");
+    background-image: url("/TEMPLATE/reampv2/public/assets/loading.gif");
```
jdeanquin-dg, line 39: use env variable PUBLIC_URL so it can load in dev

jdeanquin-dg (changes requested):

jdeanquin-dg (approved):

PR #3670, opened 2021-03-26
AMP-30039: Fix activity saving message in geocoding

vchihai:


amp/TEMPLATE/reampv2/src/modules/geocoder/actions/geocodingAction.js
```diff
@@ -72,22 +72,22 @@ export function fetchGeocodingErrorWithCode(error, status, code) {
     }
 }
 
-export function geocodeLocationPending(activity_id, location_id, accepted) {
+export function geocodeLocationPending(amp_id, location_id, accepted) {
     return {
```
jdeanquin-dg, line 76: when the object name and value is the same you can do smth like this to simplify the code
`{ amp_id,
location_id,
accepted}`

please change this in the files of this commit

jdeanquin-dg (changes requested):

jdeanquin-dg (approved):

PR #3731, opened 2021-05-15
task/AMP-30126/load report

ginchauspe:


jdeanquin-dg (commented):
it is correct that
amp/TEMPLATE/reampv2/src/modules/report_generator/actions/filterActions.js
is empty? if so please delete

jdeanquin-dg (changes requested):
see my comment

PR #3824, opened 2021-06-17
AMP-30227: Remove old gs entry.

ginchauspe:


jdeanquin-dg (changes requested):
Please remove any translation too

jdeanquin-dg (approved):

PR #3955, opened 2022-11-17
Task/amp 30363

timothymugodev:


jdeanquin-dg:
It seems that you created the branch from AMP-30368 ticket instead of develop. Please recreate the branch branching from develop if not we will be merging untested code.

jdeanquin-dg (changes requested):
see my comment

PR #3959, opened 2022-11-21
Task/30370

timothymugodev:


jdeanquin-dg:
You need to create always the branches from develop not your previous branch. if you do it that way then you may drag untested changes to develop.

jdeanquin-dg (changes requested):
see my comment.
also remember that the branch name should be task/AMP-XXXXXX/brief-description-change

timothymugodev:
Can I implement the changes in my next PR?

PR #3971, opened 2022-12-01
AMP-30408 added check to check if retlist is empty

timothymugodev:


jdeanquin-dg (approved):
to make code clearer use the attached patch.

PR #3998, opened 2023-03-01
AMP-30498 added start date and end date on the multiprogram manager

timothygachengo:


amp/WEB-INF/src/org/digijava/module/aim/action/AddTheme.java
```diff
@@ -329,6 +332,16 @@ private void fillTheme(AmpTheme ampTheme, ThemeForm themeForm) {
         ampTheme.setEnvironmentConsiderations(themeForm.getProgramEnvironmentConsiderations());
         ampTheme.setShowInRMFilters(themeForm.getShowInRMFilters());
 
+        if (themeForm.getStartDate() != null) {
+            Date castStartDate = DateConversion.getDate(themeForm.getStartDate(), "dd/MM/yyyy");
```
vchihai, line 336: It would be better to use a constant or to use the default one, by removing the format string; check DateConversion.getDate()

PR #4013, opened 2023-03-16
AMP-30409: Inconsistency in dash vs report data caused by difference …

drychter:
…in date range in global settings vs real dash. Reports were not displaying date property, setting start and end date to default


amp/TEMPLATE/ampTemplate/node_modules/amp-filter/src/views/filters-view.js
```diff
@@ -311,6 +311,7 @@ module.exports = Backbone.View.extend({
         var self = this;
         this.container = $(options.containerId);
         var serializedFilters = { filters: [] };
+        debugger;
```
jdeanquin-dg, line 314: please remove the debugger from here

jdeanquin-dg (changes requested):

amp/TEMPLATE/ampTemplate/node_modules/amp-filter/src/views/filters-view.js
drychter: good catch! done

jdeanquin-dg (dismissed):

jdeanquin-dg (approved):

PR #4017, opened 2023-03-21
Task/amp 30501/auto select dates if a program is selected

timothygachengo:


timothygachengo:
This branch contains all the changes of the indicator manager FE

PR #4024, opened 2023-04-03
AMP-30490 updated site_domain to remove default language code

timothygachengo:


jdeanquin-dg:
this pr is not needed, see my comment in jira AMP-30490

PR #4044, opened 2023-05-04
AMP-30535 created patch to fix module visibility for parent

timothygachengo:


amp/.DS_Store
jdeanquin-dg, line 1: Please don't commit this file. Add it to .gitignore

amp/deployConfigs/selected.properties
jdeanquin-dg, line 1: Please do not commit this file, so we don't generate conflicts with other branches

amp/WEB-INF/src/org/digijava/module/aim/dbentity/AmpModulesVisibility.java
jdeanquin-dg, line 1: Even though the code committed is better, it's out of the scope of this change, so please revert.

jdeanquin-dg (changes requested):

amp/deployConfigs/selected.properties
timothygachengo: Should I add this to the .gitignore?

jdeanquin-dg, line 1: you dont need to delete, you dont need to commit please revert the deletion and remove it from gitignore

jdeanquin-dg (changes requested):

jdeanquin-dg (dismissed):

jdeanquin-dg:
replaced by #4137

PR #4106, opened 2023-08-31
task/ggw-74/templates are not deleting from global fm

dynamodenis:


amp/WEB-INF/src/org/digijava/module/aim/util/FeaturesUtil.java
```diff
@@ -1254,24 +1254,32 @@ public static boolean deleteTemplateVisibility(Long id) {
...
+                hbsession.getTransaction().commit();
             }
-            ft.clearFields();
-            ft.getFeatures().clear();
-            ft.getItems().clear();
-            hbsession.delete(ft);
-            
         }
         catch (Exception ex) {
+            if (hbsession != null && hbsession.getTransaction().isActive()) {
```
jdeanquin-dg, line 1274: we should not explicitly commit or rollback since that is handled by the hibernate filter

jdeanquin-dg (changes requested):

amp/WEB-INF/src/org/digijava/module/aim/util/FeaturesUtil.java
dynamodenis: Okay noted

amp/WEB-INF/src/org/digijava/module/aim/util/FeaturesUtil.java
```diff
@@ -1271,12 +1270,10 @@ public static boolean deleteTemplateVisibility(Long id) {
...
         catch (Exception ex) {
-            if (hbsession != null && hbsession.getTransaction().isActive()) {
-                hbsession.getTransaction().rollback();
-            }
             logger.error(ex.getMessage(), ex);
             return false;
-        } finally {
+        }
+        finally {
             if (hbsession != null) {
```
jdeanquin-dg, line 1277: you don even need to close the session that is taken care by the filter while committing

jdeanquin-dg (changes requested):

jdeanquin-dg (dismissed):

PR #4121, opened 2023-09-22
GGW-91 fix change password bug

timothygachengo:


amp/WEB-INF/src/org/digijava/module/um/util/DbUtil.java
```diff
@@ -234,23 +234,14 @@ public static void updatePassword(String user, String oldPassword,
...
-            session.save(iterUser);
+            User foundUser = UserUtils.getUserByEmailAddress(user);
 
-            //tx.commit();
+            foundUser.setPassword(ShaCrypt.crypt(newPassword.trim()).trim());
+            foundUser.setSalt(Long.toString(newPassword.trim().hashCode()));
+            foundUser.updateLastModified();
+
+            session.update(foundUser);
+            session.getTransaction().commit();
```
jdeanquin-dg, line 244: there is no need to commit this is handled by the hibernate intercetor

jdeanquin-dg (changes requested):

jdeanquin-dg:
This code does not fix the problem. Closing

PR #4130, opened 2023-10-06
Changing amp-ui reference

drossi-dg:


jdeanquin-dg (changes requested):
you need to delete your local package-lock.json 
run nom install and commit package-lock if not its still points to the old branch

jdeanquin-dg (approved):

PR #4145, opened 2023-10-12
Issue/amp 30735/ethiopia  error when you add region column to report

drossi-dg:
Improving error message


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/reports/designer/ReportColumnProvider.java
```diff
@@ -139,7 +139,7 @@ public class ReportColumnProvider extends ReportEntityProvider {
     private static final Comparator<AmpColumns> ME_COLS_COMPARATOR =
             Comparator.comparing(c -> ME_COLUMNS_ORDER.getOrDefault(c.getColumnName(), DEFAULT_ORDER_VALUE));
 
-    private TranslatorService translatorService;
+    public TranslatorService translatorService;
```
jdeanquin-dg: to keep the same structure and since translator service is a singleton instead of making it punlaick create a variable inside ReportColumValidator and populate it at creation time

jdeanquin-dg (changes requested):

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/reports/designer/ReportColumnProvider.java
drossi-dg: done

PR #4188, opened 2023-12-04
GGW-139 created created report endpoint to get chart data

timothygachengo:


jdeanquin-dg (changes requested):
we already have this endpoint that does this
/rest/data/report/custom/paginate

PR #4199, opened 2023-12-19
Feature/AMP-30574/sectors mapping

drossi-dg:


amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/gpi/ValidationUtils.java
```diff
@@ -34,6 +31,17 @@ public static void requireValid(Object obj) {
         }
     }
 
+    public static void valuesValid(Collection possibleValues, Object value) {
```
brianbrix, line 34: Can we add something like `if (possibleValues == null) {
        throw new IllegalArgumentException("possibleValues cannot be null");
    }` to avoid a nullpointer exception

amp/WEB-INF/src/org/digijava/module/aim/util/SectorUtil.java
```diff
@@ -1214,12 +1238,109 @@ public static List<AmpActivityVersion> getActivitiesForSector(Long id) {
...
+        Session session = null;
+        Collection col = null;
+        Query qry = null;
+
+        try {
+            session = PersistenceManager.getSession();
+            queryString = "select asm from " + AmpSectorMapping.class.getName() + " asm";
+            qry = session.createQuery(queryString);
+            col = qry.list();
+            session.flush();
```
brianbrix: No need to flush after select

amp/WEB-INF/src/org/digijava/module/aim/util/SectorUtil.java
```diff
@@ -1214,12 +1238,109 @@ public static List<AmpActivityVersion> getActivitiesForSector(Long id) {
...
+            session = PersistenceManager.getRequestDBSession();
+
+            for (AmpSectorMapping em : existingMappings) {
+                session.delete(em);
+            }
+            session.flush();
+            for (AmpSectorMapping asm : sectorMappings) {
+                session.save(asm);
+            }
+            session.flush();
```
brianbrix, line 1319: We can combine the flush operations to one at the end

amp/WEB-INF/src/org/digijava/module/aim/util/SectorUtil.java
drossi-dg: I added flush here because without flush It didn't work sometimes

amp/WEB-INF/src/org/digijava/module/aim/util/SectorUtil.java
drossi-dg: I made it but it didn't work. When I used only one flush at the end, I receive SQL error by dyplicate rows, but with 2 flush the deletion was made succefully and the new mappings were saved ok

amp/WEB-INF/src/org/digijava/module/aim/util/SectorUtil.java
brianbrix: Okay...

amp/WEB-INF/src/org/digijava/module/aim/util/SectorUtil.java
brianbrix: Okay

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/gpi/ValidationUtils.java
drossi-dg: I will

amp/WEB-INF/src/org/digijava/module/aim/util/SectorUtil.java
brianbrix: I was wondering what get if we dont flush after a select operation. Because typically we should not flush after a select since there is no use for synchronizing the persistence context.. I think we should check this and rectify it.

drossi-dg: done

amp/WEB-INF/src/org/digijava/kernel/ampapi/endpoints/gpi/ValidationUtils.java
drossi-dg: done

jdeanquin-dg (dismissed):

PR #4255, opened 2024-03-07
AMP-30832 fix activity fields cloning

jdeanquin-dg:


amp/WEB-INF/src/org/digijava/kernel/user/User.java
```diff
@@ -84,7 +84,9 @@ public class User
     private Set<AmpOrganisation> assignedOrgs;
     private Date passwordChangedAt;
 
-    public User() {}
+    public User() {
+        System.out.println("hola");
```
brianbrix: Better to use a logger if we need to log this.

brianbrix (dismissed):

brianbrix (dismissed):

brianbrix (dismissed):

brianbrix (dismissed):

brianbrix (dismissed):

PR #4291, opened 2024-04-10
Fix/ggw 297/filter fails to load in management workspace

brianbrix:


jdeanquin-dg:
please create a branch that only commits the changes for GGW-297

PR #4315, opened 2024-05-20
Feature/amp 30876/weird behaviour of hide unhide location item

brianbrix:


jdeanquin-dg (changes requested):
this goes to #future/GGW

PR #4377, opened 2025-03-12
Indirect program issue

brianbrix:


amp/WEB-INF/src/org/dgfoundation/amp/onepager/util/IndirectProgramUpdater.java
```diff
@@ -74,8 +74,7 @@ private void includeAncestors(Map<AmpTheme, Set<AmpTheme>> mapping) {
             if (srcParentProgram != null) {
                 Set<AmpTheme> dstParentPrograms = mapping.computeIfAbsent(srcParentProgram, p -> newSetComparingById());
 
-                mapping.getOrDefault(srcProgram, emptySet()).stream()
-                        .map(AmpTheme::getParentThemeId)
```
brianbrix, line 78: Instead of adding the dstProgram's parent to the collection we add the dstProgram itself..

PR #4393, opened 2025-05-08
Imported database issues

brianbrix:


amp/src/main/resources/xmlpatches/4.0/Update-Import-Tables-v3.xml
jdeanquin-dg, line 1: this needs to be executed manually and not as a patch because this patch runs on every installation

jdeanquin-dg (changes requested):

jdeanquin-dg (approved):

PR #4439, opened 2025-10-14
AMP-31057 :Import the AMP data into WP API

brianbrix:


amp/src/main/resources/xmlpatches/4.0/AMP-31057-Import-Donor-Funding-Job.xml
```diff
@@ -0,0 +1,23 @@
...
+        <condition type="entryInTableMissing" tablename="amp_quartz_job_class"
+                   columnname="jc_name" columnvalue="Import Donor funding job" />
+    </trigger>
+    <apply>
+        <script>
+            <lang delimiter=";" type="postgres">
+                INSERT INTO amp_quartz_job_class (jc_id,jc_name, jc_class_fullname)
+                VALUES(nextval('amp_quartz_job_class_seq'),'Import Donor funding job',
+                'org.digijava.module.message.jobs.AmpDonorFundingJob');
+            </lang>
```
jdeanquin-dg, line 19: this patch will fail if AmpDonorFundingJob already exists

jdeanquin-dg (changes requested):

amp/src/main/resources/xmlpatches/4.0/AMP-31057-Import-Donor-Funding-Job.xml
jdeanquin-dg: its already taken care, I missed it

PR #4444, opened 2025-10-27
AMP-31057 :  Trying resolve for missing responsible org

brianbrix:


jdeanquin-dg (changes requested):
why are you changing the FM paths for activity fields?

brianbrix:
Reverted to initial correct FM Paths

jdeanquin-dg (approved):

PR #4467, opened 2026-05-12
AMP-31111: issue- adding a new tab cause error when a region column added

brianbrix:


brianbrix:
The error says "Invalid Column(9)" ...when you check in the database column 9 is Adminstrative Level 1
For Regional Report type .. the LOCATION_ADM_LEVEL_1 was being hidden(Chccke ReportColumnProvider class line number 295) which meant that if passed to the report it could be rejected ..The frontend sends it so what we do is: Iterate through the request's columns list and swap out any occurrence of LOCATION_ADM_LEVEL_1's ID with REGIONAL_REGION's ID

jdeanquin-dg (approved):

